PR #9, opened 2024-12-06
Add SES and SQS client

ehanson8:
This code is largely copied over from `wiley-deposits` with some minor changes to align it with our current coding practices.

Not possible yet

YES

NO

- https://mitlibraries.atlassian.net/browse/IN-1099

### Developer

### Code Reviewer(s)


dsc/sqs.py
```diff
@@ -0,0 +1,203 @@
...
+        """
+        valid = False
+        if (
+            self.valid_result_message_attributes(sqs_message=sqs_message)
+            and self.valid_result_message_body(sqs_message=sqs_message)
+            and sqs_message.get("ReceiptHandle")
+        ):
+            valid = True
+        else:
+            logger.exception("Failed to parse SQS message: %s", sqs_message)
```
ghukill, line 202: Going to pick on this `logger.exception` line, but I think it may apply more broadly to the flow here, and the logging messages from the helper methods.

Let's say that `self.valid_result_message_body()` cannot load the `Body` part of the message as JSON.  Wouldn't both that method, and this method `valid_sqs_message()` (or `validate_sqs_message()` if renamed) both log the full SQS message?  Something like:

```
Failed to parse SQS message body: <SQS message here...>
Failed to parse SQS message: <SQS message here...>
```

To me what feels ideal would be logging like:

```
"Invalid SQS message: failed to parse message body
```

Where the first part is coming from the `validate_sqs_message()` method, but the _reason_ "failed to parse message Body as JSON" came from the helper method `valid_result_message_body`.

It's all pretty minor, and frankly not blocking for my approval, but feels like maybe those helper methods that are doing the validation work could just return tuples like `(<success:bool>, <reason:str>)` that this `validate_sqs_message()` method could use.

ghukill: I should have led with: I do like this validation approach!  It's nice they are in standalone methods.  

I'd be okay with the approach as-is, there are things I like about it too.  Please take all comments here as optional / conversational, but by no means requested changes.

ghukill (approved):
Looking good!  I did leave some comments about SQS validation, but wholly optional.  I believe it would be functional and understandable as-is.

dsc/sqs.py
```diff
@@ -0,0 +1,203 @@
...
+            sqs_message: An SQS message to be evaluated.
+        """
+        valid = False
+        if "Body" in sqs_message and json.loads(str(sqs_message["Body"])):
+            valid = True
+        else:
+            logger.exception("Failed to parse SQS message body: %s", sqs_message)
+        return valid
+
+    def valid_sqs_message(self, sqs_message: MessageTypeDef) -> bool:
```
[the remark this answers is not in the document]
ehanson8: Very good call, I'll update this!

dsc/sqs.py
```diff
@@ -0,0 +1,203 @@
...
+        EmptyResponseMetadataTypeDef,
+        MessageAttributeValueTypeDef,
+        MessageTypeDef,
+        SendMessageResultTypeDef,
+    )
+
+logger = logging.getLogger(__name__)
+
+
+class SQSClient:
```
jonavellecuerdo, line 24: Commenting up here but applies to all `logger` calls in the client: Can you update the log calls to use f-string?

jonavellecuerdo (changes requested):
A couple change requests! Let me know what you think. Just had a few thoughts on what might be good updates when looking at clients in our more recent projects. 🤓

dsc/ses.py
```diff
@@ -0,0 +1,84 @@
...
+        Args:
+           subject: The subject of the email.
+           attachment_content: The content of the email attachment.
+           attachment_name: The name of the email attachment.
+           source_email_address: The email address of the sender.
+           recipient_email_address: The email address of the receipient.
+        """
+        message = self._create_email(subject, attachment_content, attachment_name)
+        self._send_email(source_email_address, recipient_email_address, message)
+        logger.debug("Logs sent to %s", recipient_email_address)
```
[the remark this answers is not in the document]
ehanson8: Yes, I already caught this as I was doing some refactoring in response to Graham's comments so in progress!

dsc/sqs.py
ehanson8: I did refactor this method to prevent dupe logging, let me know how it looks

ghukill: @ehanson8 - maybe missing it somehow, but not seeing a new commit?

dsc/sqs.py
```diff
@@ -0,0 +1,203 @@
...
+    """A class to perform common SQS operations for this application."""
+
+    def __init__(self, region: str, base_url: str, queue_name: str) -> None:
+        self.client = client("sqs", region_name=region)
+        self.base_url: str = base_url
+        self.queue_name: str = queue_name
```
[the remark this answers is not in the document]
ehanson8: Good call, updated!

dsc/sqs.py
ehanson8: Just pushed!

dsc/utilities/aws/sqs.py
```diff
@@ -139,17 +151,35 @@ def send(
...
-            QueueUrl=f"{self.base_url}{self.queue_name}",
+            QueueUrl=self.queue_url,
             MessageAttributes=message_attributes,
             MessageBody=str(message_body),
         )
-        logger.debug("Response from SQS queue: %s", response)
+        logger.debug(f"Response from SQS queue: {response}")
         return response
 
+    def validate_message(self, sqs_message: MessageTypeDef) -> bool:
```
ghukill, line 163: I know that I submitted an approval already, and I do think this current approach will successfully validate messages, but taking another pass at the updates I do wonder if there is an opportunity to streamline some of this validation even more.  I would imagine that _much_ of this application will be about validation, so some patterns to follow throughout could be helpful to establish early.

I see that all this validation is under the umbrella of [this call to self.validate_message()](https://github.com/MITLibraries/dspace-submission-composer/blob/IN-1099-ses-and-sqs-clients/dsc/utilities/aws/sqs.py#L117-L118) in `self.process_result_message()`:

```python
if not self.validate_message(sqs_message):
    raise InvalidSQSMessageError
```

Where the ultimate goal is to a) log what happend, and b) raise `InvalidSQSMessageError` from this method.

What if each validation helper method was responsible for directly raising `InvalidSQSMessageError` if they found something wrong?  and `self.validate_message()` is just an orchestrator of calling these methods?  Something along the lines of:

```python
def self.validate_message() -> None:
    # call all validators here
    # if all pass, return None, and all is well!
    # if any raise an exception, it will bubble up from here and reach self.process_result_message()
    #   where this was originally called from

def validate_message_attributes() -> None:
   # each validator returns None
   # do checks, raise InvalidSQSMessageError with a meaningful message if something wrong

# continue to define validators as needed
```

I think this leans into the EAFP (Easier to ask for forgiveness than permission) python pattern.  If we call `self.validate_message()` and nothing happens, great.  But if any of those validator methods raise that `InvalidSQSMessageError` exception, it will:

1. get raised from that lowest validator method
2. bubble up through `validate_message()`
3. bubble up from `process_results_message()` [here](https://github.com/MITLibraries/dspace-submission-composer/blob/IN-1099-ses-and-sqs-clients/dsc/utilities/aws/sqs.py#L117) just as it does now

This pattern is extensible, becuase you can just keep calling validator methods, knowing they are responsible for raising that exception.

To reiterate: with only two validations here, I think the current approach works as-is.  But I have this hunch that this DSC app will be validation heavy, and a validation pattern that is applied somewhat globally could be helpful.

jonavellecuerdo: I like @ghukill 's proposal to have `validate_message` simply call validation methods that are responsible for raising `InvalidSQSMessageError` with meaningful messages.

jonavellecuerdo (commented):
Thank you for making these additional changes! Approved but waiting on your thoughts re: implementing @ghukill 's proposal for validation methods approach.

dsc/utilities/aws/sqs.py
ehanson8: Thank you, this is definitely a better approach, I greatly appreciate the suggestion and I'll rework!

ehanson8:
@ghukill @jonavellecuerdo Updated the validation methods!

dsc/utilities/aws/sqs.py
```diff
@@ -0,0 +1,203 @@
...
+
+        return response
+
+    def process_result_message(self, sqs_message: MessageTypeDef) -> tuple[str, str]:
+        """Validate, extract data, and delete an SQS result message.
+
+        Args:
+            sqs_message: An SQS result message to be processed.
+        """
+        self.validate_message(sqs_message)
```
ghukill, line 117: 🔥

dsc/utilities/aws/sqs.py
```diff
@@ -0,0 +1,203 @@
...
+        Args:
+            sqs_message: An SQS message to be evaluated.
+        """
+        if (
+            "MessageAttributes" not in sqs_message
+            or not any(
+                field
+                for field in sqs_message["MessageAttributes"]
+                if "PackageID" in field
+            )
```
ghukill, line 188: I must admint, this one is still tripping me up.  Are we looking for `"PackageID"` as a key in the `sqs_message["MessageAttributes"]` dictionary?  

If so, could we rewrite?

```python
or "PackageID" not in sqs_message["MessageAttributes"]
```

jonavellecuerdo: Now that we are looking at the validation method even closer, I am curious about this as well. Debugging the test, it seems the an `sqs_message` looks like:

```text
{
    "MessageId": "47179327-6a75-467d-b431-d7bfd802b15e",
    "ReceiptHandle": "zqaejnmrlfkzlxsxexmzsnyjsiugkotjdouweeephwuigifrlsrwxshfqmguyjttklxxrogfiohgwezhqpywvrbgbueebxlmuenrwtgtgmhbaugtywthkppbixxbjowmkqcwiqapxjpirgeeuowpgdkrdskhawgimaaynsbekemlfxoapaskuewaa",
    "MD5OfBody": "2affecad36e208b6be933fce09b7a89e",
    "Body": '{"ResultType": "success", "ItemHandle": "1721.1/131022", "lastModified": "Thu Sep 09 17:56:39 UTC 2021", "Bitstreams": [{"BitstreamName": "10.1002-term.3131.pdf", "BitstreamUUID": "a1b2c3d4e5", "BitstreamChecksum": {"value": "a4e0f4930dfaff904fa3c6c85b0b8ecc", "checkSumAlgorithm": "MD5"}}]}',
    "MD5OfMessageAttributes": "5edd88184e055125bdab8e1b116ecf5e",
    "MessageAttributes": {
        "PackageID": {"StringValue": "10.1002/term.3131", "DataType": "String"},
        "SubmissionSource": {"StringValue": "Submission system", "DataType": "String"},
    },
}
```

jonavellecuerdo (commented):
Just two minor change requests, including @ghukill 's suggestion!

dsc/utilities/aws/sqs.py
ehanson8: Don't know how I came up with that check and how it survived this long without refactoring but yes `or "PackageID" not in sqs_message["MessageAttributes"]` is much simpler

dsc/utilities/aws/sqs.py
```diff
@@ -0,0 +1,202 @@
...
+    @staticmethod
+    def validate_message_attributes(sqs_message: MessageTypeDef) -> None:
+        """Validate that "MessageAttributes" field is formatted as expected.
+
+        Args:
+            sqs_message: An SQS message to be evaluated.
+        """
+        if (
+            "MessageAttributes" not in sqs_message
+            or "PackageID" not in sqs_message["MessageAttributes"]
```
ghukill, line 185: 👍

jonavellecuerdo (approved):

PR #21, opened 2024-12-13
In 1101 base workflow

ehanson8:
Certain code blocks have bounced around while writing this and I am finding there many different ways to organize the boundaries between the CLI command, `BaseWorkflow` class, and the `ItemSubmission` class. The `abstractmethod`s on the `BaseWorkflow` class outline the boundary I'm seeing between the workflow  and the item, with the expectation that unwritten CLI commands will be the interface with the DSS SQS input/output queues and the SES emails. My guiding principle was to minimize the number of values passed between various classes, but as always, this is wide open to iteration! 

Not possible yet

YES

NO

- https://mitlibraries.atlassian.net/browse/IN-1101

### Developer

### Code Reviewer(s)


dsc/item_submission.py
```diff
@@ -0,0 +1,68 @@
...
+        """
+        dspace_metadata = self.create_dspace_metadata()
+        self.s3_client.put_file(
+            json.dumps(dspace_metadata),
+            bucket,
+            self.metadata_keyname,
+        )
+        metadata_uri = f"s3://{bucket}/{self.metadata_keyname}"
+        logger.info(f"Metadata uploaded to S3: {metadata_uri}")
+        self.metadata_uri = metadata_uri
```
ehanson8, line 36: I considered exception handling here but as with `wiley-deposits`, I think it's better to crash the run on a `ClientError` than continue if we can't get the metadata file into S3

dsc/base.py
```diff
@@ -0,0 +1,96 @@
...
+        """
+        s3_client = S3Client()
+        batch_metadata = self.get_batch_metadata()
+        for item_metadata in batch_metadata:
+            item_identifier = self.get_item_identifier(item_metadata)
+            logger.info(f"Processing submission for '{item_identifier}'")
+            metadata_keyname = f"{self.s3_prefix}/{item_identifier}_metadata.json"
+            item_submission = ItemSubmission(
+                source_metadata=item_metadata,
+                metadata_mapping=self.metadata_mapping,
```
ghukill, line 56: My next can-o-worms question is here.

So it's safe to assume that `metadata_mapping` is a pretty critically important part of each workflow?  This would be a dictionary that maps the item's "original" metadata dictionary to metadata fields that align with the dspace metadata profile?  

Super rough example:

```python
{
    "title":"dc.title",
    ...
}
```

Do you expect there will be any difficulties with nesting, logic, or multivalued fields?

For example: what if the input metadata for workflow `Foo` has a column in the CSV with semicolon delimited subjects, e.g. `apples;washington;mountains`.  And the goal would be multiple `dc.subject` field values for each.  I'm unsure how a dictionary mapping could explode a single value into multiple values.

But, this assumes quite a bit.  First, that DSpace's metadata profile can (or will) allow for more complex or multivalued fields.  Second, that we don't just require the input metadata format to have basically a 1:1 mapping to DSpace fields already present.

I think this consideration points to `ItemSubmission.create_dspace_metadata()` (which I have not yet looked closely at) which I would expect to take the `source_metadata` and apply the `metadata_mapping` to get the final item metadata for DSpace.

Going to leave this comment here, as-is, and followup in that method.

dsc/item_submission.py
```diff
@@ -0,0 +1,68 @@
...
+
+    def create_dspace_metadata(self) -> dict[str, Any]:
+        """Create DSpace metadata from the item's source metadata."""
+        metadata_entries = []
+        for field_name, field_mapping in self.metadata_mapping.items():
+            if field_name not in ["item_identifier", "source_system_identifier"]:
+
+                field_value = self.source_metadata.get(field_mapping["source_field_name"])
+                if field_value:
+                    delimiter = field_mapping["delimiter"]
```
ghukill, line 46: Ah-ha!  Related to my comment above about `metadata_mapping`, I see now that each entry in that is an object itself, which can provide more information for this mapping.  

**Request**: can you add an example of what this metadata mapping might look like?  With an example, I think I would understand that a single dictionary could suffice.

tests/conftest.py
```diff
@@ -22,15 +23,51 @@ def _test_env(monkeypatch):
...
+        },
+        "dc.title": {
+            "source_field_name": "title",
+            "language": "en_US",
+            "delimiter": "",
+        },
+        "dc.contributor": {
+            "source_field_name": "contributor",
+            "language": None,
+            "delimiter": "|",
```
ghukill, line 61: Thanks for adding an example here of a delimiter.  My comment above requesting an example in the docstring stands, but double-helpful to have this example here (and test it, of course).

dsc/base.py
ghukill, line 1: Commenting here, but this touches on file organization more generally.

- what if `base.py` was renamed to `workflows.py`, where it may contain additional classes or helper functions as the project grows
- what if `item_submission.py` was renamed to `items.py`, where it's a bit more generic to just providing classes about "Items" (maybe starting with `ItemSubmission`, but perhaps there is room eventually for something like `ItemSource` or `ItemMetadata`)
- create a folder called `/sources` where actual workflows that extend these could live

Resulting in a structure like:

```
/ dsc
  / sources
    - wiley.py  # which would contain a class like `WileyWorkflow`
    - ...
    - ...
  - items.py
  - workflows.py
```

This could help make clear that anything under `/sources` is an extension of these base files and classes.  And then each of those "base" files would be descriptive about what kind of classes it held.

NOTE: we could create the `/sources` directory now and include a file like `.gitkeep` which would ensure it gets created and committed, even though we don't have sources yet.

dsc/base.py
```diff
@@ -0,0 +1,96 @@
...
+        self.s3_bucket: str = s3_bucket
+        self.s3_prefix: str | None = s3_prefix
+
+    @final
+    def generate_submission_batch(self) -> Iterator[tuple[str, list[str]]]:
+        """Generate a batch of item submissions for the DSpace Submission Service.
+
+        MUST NOT be overridden by workflow subclasses.
+        """
+        s3_client = S3Client()
```
ghukill, line 48: I might propose that `ItemSubmission` could initialize it's own `S3Client` to use.  I know we've touched on caching and reusing clients, but I think the performance gain is basically non-existent here, given how quickly this is initialized.

I would think that `ItemSubmission.generate_and_upload_dspace_metadata()` could just instantiate and use a client right when it needs it, instead of applying to the `ItemSubmission` class instance.

Or, have a property like this if multiple methods may eventually require an S3 client:

```
@property
def s3_client(self):
    return S3Client()
```

ghukill (commented):
Submitting "Comment" as I have some questions and comments sprinkled through.

Overall, I'm definitely understanding the design.  And in these early stages, I think it's nice that it's relatively simple.

At this early stage, what I'm most interested in are the relationships between a `Workflow` and an `ItemSubmission`.  I left some comments about typings, and how it _seems_ like you're anticipating that every source record is a `dict`.  If so, I left some comments / requests to lean further into that direction.

I can't shake the feeling that perhaps some kind of class like `SourceItem` or `ItemMetadata` could be helpful, instead of passing around dictionaries... but maybe wait until you've had time to respond to these first round of questions.

dsc/item_submission.py
```diff
@@ -0,0 +1,68 @@
...
+from dataclasses import dataclass
+from typing import Any
+
+from dsc.utilities.aws.s3 import S3Client
+
+logger = logging.getLogger(__name__)
+
+
+@dataclass
+class ItemSubmission:
```
ghukill, line 12: Do you think `ItemSubmission` would benefit from some kind of `validate()` method?  I'm unsure offhand what that might look for, but I'm noticing now that we call `generate_and_upload_dspace_metadata()` in the `for` loop of the workflow and it might be helpful to check things look good before we fire it off.

dsc/base.py
```diff
@@ -0,0 +1,96 @@
...
+            s3_prefix: The S3 prefix used for objects in this workflow. This prefix does
+            NOT include the bucket name.
+        """
+        self.email_recipients: list[str] = email_recipients
+        self.metadata_mapping: dict = metadata_mapping
+        self.s3_bucket: str = s3_bucket
+        self.s3_prefix: str | None = s3_prefix
+
+    @final
+    def generate_submission_batch(self) -> Iterator[tuple[str, list[str]]]:
```
ghukill, line 43: I'm realizing after my first pass that I have a couple of fundamental questions about this base workflow method.

When this is complete, is it correct that:
1. each `ItemSubmission` instantiated in the loop will have uploaded a JSON file to S3 that is _kind_ of like a SIP for DSS to retrieve and ingest into DSpace
2. it yields the URI of each new `ItemSubmission`, which I believe to be the S3 URI of the JSON file uploaded to S3 for that item
3. it also yields `ItemSubmission.bitstream_uris`,  which are _not_ yet getting set on `ItemSubmission`

My question: what is utilizing the results of this yielding?  Will it be another class method?  or the CLI itself?  Unless I'm grossly misunderstanding this application, wouldn't something need to send these as an SQS message to the DSS input queue?

Related question -- and apologies if this will come later -- what happens with the bistream URIs?  Are they included in the SQS message for DSS?

tests/conftest.py
ehanson8: Yes, I'll add that to the docstring!

dsc/base.py
ehanson8: The yielded URIs will be utilized by the `deposit` command to generate the SQS message, I mentioned in the PR description 

> with the expectation that unwritten CLI commands will be the interface with the DSS SQS input/output queues and the SES emails.

The workflow class is intended for just the workflow-opinionated code while the CLI command will be a wrapper with universal functionality (e.g. sending SQS messages, SES emails)

And yes, the bitstream URIs are in the the SQS message as well as the URI of the JSON metadata file that is uploaded to S3, [DSS submission message spec](https://github.com/MITLibraries/dspace-submission-service/blob/main/docs/specifications/submission-message-specification.md) for reference.

The steps would be:
1. `ItemSubmission` instantiated with source metadata and bitstreams URIs
2. DSpace metadata generated from source metadata
3. DSpace metadata uploaded to S3 as JSON file and URI is returned
4. ItemSubmission yields `metadata_uris` and `bitstream_uris to `deposit` CLI command for SQS message
5. `deposit` CLI command generates and send SQS message to DSS

dsc/base.py
ehanson8: Will do!

dsc/item_submission.py
ehanson8: I did have a `valid_dspace_metadata` method in `wiley-deposits` but I may want to rework it since that had specific field names to validate while these workflows may be less constrained. But yes, there should be some validation!

dsc/base.py
ghukill: Ah, sorry missed that in the PR!  Thanks for pointing that out.

And thanks for sharing the DSS submission message spec.  That document is SUPER helpful, and I think it'd be handy to include in docstrings in _this_ application where applicable.

_And_, thanks for the steps; I had missed that `get_bitstream_uris()` is a method that workflows will need to define, and that's where they come from.

This is quite helpful for helping my clarify my next question: any reason why we couldn't yield the `ItemSubmission` object itself?  If I'm understanding correctly, it will have the `metadata_uri` and the `bitstream_uris`.

Happy to wait on that until the CLI level orchestration is in place, where these would get sent out as SQS messages.  Are you envisioning some kind of `DSSInputMessage` class that could encapsulate preparing the final SQS payload for sending, that would meet the [spec](https://github.com/MITLibraries/dspace-submission-service/blob/main/docs/specifications/submission-message-specification.md)?

Which I think has a natural follow-up question: why couldn't `ItemSubmission` have such a method to perform this work?  Like `to_dss_input_queue_message()`?

Happy to meet and discuss if helpful, or keep hashing out in comments.  When I see a class like `ItemSubmission`, which is responsible for uploading the components of a SIP to S3 (metadata and files), I wonder why it wouldn't have the capabilities of actually preparing (and maybe actually sending?) the SQS message for DSS.

dsc/base.py
```diff
@@ -0,0 +1,96 @@
...
+        """Generate a batch of item submissions for the DSpace Submission Service.
+
+        MUST NOT be overridden by workflow subclasses.
+        """
+        s3_client = S3Client()
+        batch_metadata = self.get_batch_metadata()
+        for item_metadata in batch_metadata:
+            item_identifier = self.get_item_identifier(item_metadata)
```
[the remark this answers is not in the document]
ehanson8: I would say every method the the workflow classes are tightly coupled and opinionated to that workflow. In retrospect, I thought I errored in typing `item_metadata=Any` in `get_item_identifier` and `source_metadata=dict[str, Any]` in `ItemSubmission` since I didn't expect that object to be transformed between getting the identifier and instantiating `ItemSubmission`. 

However! That is now suggesting to me that a `get_item_metadata(batch_metadata)` method is needed and that those type hints should remain:

```
batch_metadata = self.get_batch_metadata()
for item_metadata in self.get_item_metadata(batch_metadata):
      item_identifier = self.get_item_identifier(item_metadata)
```

Currently, `batch_metadata` in most workflows is a CSV so `get_item_metadata` would be a simple `dict` iterator.  For `OpenCourseWare`, it's a list of zip files, so `get_item_metadata` iterates and extracts a `dict` from each zip file. If a future source uses `XML`, `get_item_metadata` would iterate and generate a `dict` from the `XML`. Thoughts?

dsc/base.py
ehanson8: > Which I think has a natural follow-up question: why couldn't `ItemSubmission` have such a method to perform this work? Like `to_dss_input_queue_message()`?

I think it certainly could! Let's get @jonavellecuerdo 's thoughts as well but now I'm inclined to build SQS message formatting  into `ItemSubmission` rather than the CLI command

dsc/base.py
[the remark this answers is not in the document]
ghukill: So would it be accurate to say?

1. `get_batch_metadata()` --> retrieves metadata from somewhere, but what is returned still contains multiple items; probably most often downloading from S3, but theoretically could get it other ways too (e.g. an HTTP request to ETD... 😎 ).  The result might be a single CSV reader/dataframe, a single parsed XML file object, etc.

2. `get_item_metadata()` --> would yield actual dictionaries of item metadata, for each item, doing whatever is needed (e.g. unzipping files, parsing XML, looping through CSV rows, etc.)

I'm having trouble articulating what feels confusing about the relationship of,
- `get_batch_metadata`
- `get_item_identifier`
- `get_bitstream_uris`
- and now the proposed `get_item_metadata`

It feels kind of like they are very tightly coupled, and creating new workflows that extend this class would be jumping through hoops to define those inter-related methods.  When -- and maybe missing something here -- a workflow could just be responsible for defining something like `fetch_items()` which would yield nearly everything you need to instantiate an `ItemSubmission`.  

Which, and I'm feeling dizzy from circular questioning, why wouldn't they _directly_ yield an `ItemSubmission` then?  

Each of those workflows could, and _should_, define helper methods as needed to support this work.  But in some cases, like reading a CSV from S3, it feels like `get_batch_metadata` (reading the CSV) and yielding items `get_item_metadata` could be achieved with a couple lines like:

```python
import pandas as pd

df = pd.read_csv("s3://....")
for _, row in df.iterrows():
   yield row # <----- basically an Item
```

Sorry if confusing the issue, and happy to hold off on these questions until more is established.

dsc/base.py
jonavellecuerdo: I do think yielding the actual `ItemSubmission` instance will allow the object to be accessed outside the context of the `BaseWorkflow.generate_submission_batch` method. 🤔 I also think if all `ItemSubmission` instances must comply with the DSS submission message spec, then--as @ehanson8 writes--it also follows that defining the SQS message formatting in `ItemSubmission` would make sense!

dsc/item_submission.py
jonavellecuerdo: Side note: If we find that `dataclass` module is the right fit for this module, I propose calling the proposed `validate()` method in a `__post_init__` method!

dsc/workflows/base/base_workflow.py
ghukill, line 1: Very minor, very optional: class `BaseWorkflow` could be defined in the `dsc/workflows/base/__init__.py` file if we wanted to avoid another file like `base_workflow.py` or `base.py`.

dsc/workflows/base/base_workflow.py
```diff
@@ -0,0 +1,199 @@
...
+        s3_prefix: str | None,
+        collection_handle: str,
+        output_queue: str,
+    ) -> None:
+        """Initialize base instance.
+
+        Args:
+            workflow_name: The name of the workflow.
+            submission_system: The system to which item submissions will be sent (e.g.
+            DSpace@MIT)
```
ghukill, line 35: This reveals some ignorance about the greater context of the application, but when would it _not_ be `DSpace@MIT`?  For Dome?  

Would it make sense to make this only values from an `Enum`?  Is this used directly in the SQS message that then tells DSS something special?

Or... wonder if it would make sense for each workflow defined to define this on the class itself (i.e. a class level attribute, [here's an example from GeoHarvester](https://github.com/MITLibraries/geo-harvester/blob/main/harvester/records/sources/alma.py#L27))?  Instead of passing to the Workflow when instantiated.  Seems like any given workflow will only ever want to deposit into a single repository.

ghukill: If class level (aka workflow level) static attributes might make sense, do wonder about other things passed here.  For example: would `output_queue` also be pretty workflow specific?  or `collection_handle`?  I honestly don't know, but maybe worth considering.

dsc/workflows/base/base_workflow.py
```diff
@@ -0,0 +1,199 @@
...
+    @final
+    def item_submissions_iter(self) -> Iterator[ItemSubmission]:
+        """Generate a batch of item submissions for the DSpace Submission Service.
+
+        MUST NOT be overridden by workflow subclasses.
+        """
+        for item_metadata in self.batch_metadata_iter():
+            item_identifier = self.get_item_identifier(item_metadata)
+            logger.info(f"Processing submission for '{item_identifier}'")
+            metadata_keyname = f"{self.s3_prefix}/{item_identifier}_metadata.json"
```
ghukill, line 88: Optional: I found myself tripping on `metadata_keyname` for some reason.  Maybe something like `metadata_s3_key`?  At a glance, I would assume that's everything _but_ the bucket.  Has some nice symmetry with `s3_bucket` and `s3_prefix`.  

And maybe this suggests the attribute could be `ItemSubmission.metadata_s3_key`.

dsc/workflows/base/base_workflow.py
```diff
@@ -0,0 +1,199 @@
...
+            dspace_metadata = self.create_dspace_metadata(item_metadata)
+            if self.validate_dspace_metadata(dspace_metadata):
+                item_submission = ItemSubmission(
+                    dspace_metadata=dspace_metadata,
+                    bitstream_uris=self.get_bitstream_uris(item_identifier),
+                    metadata_keyname=metadata_keyname,
+                )
+                yield item_submission
```
ghukill, line 96: An "exit early" pattern could be:

```python
self.validate_dspace_metadata(dspace_metadata) # if fails validation it raise Exception
item_submission = ... # rest of code
```

This keeps the indenting low, and I _think_ might match some validation approaches elsewhere (try it, if good, does nothing).

dsc/workflows/base/base_workflow.py
```diff
@@ -0,0 +1,199 @@
...
+        """Get identifier for an item submission according to the workflow subclass.
+
+        MUST be overridden by workflow subclasses.
+
+        Args:
+            item_metadata: The item metadata from which the item identifier is extracted.
+        """
+
+    @final
+    def create_dspace_metadata(self, item_metadata: dict[str, Any]) -> dict[str, Any]:
```
ghukill, line 116: If this pattern is flexible enough for all workflows, I think this could be a real strength of this application: requiring basically a "raw" item dictionary + workflow mapping = final dspace metadata.  If this is truly the only transformation logic needed, across all workflows, that would be pretty great!

dsc/workflows/base/base_workflow.py
```diff
@@ -0,0 +1,199 @@
...
+        }
+
+        MUST NOT be overridden by workflow subclasses.
+
+        Args:
+            item_metadata: Item metadata from which the DSpace metadata will be derived.
+        """
+        metadata_entries = []
+        for field_name, field_mapping in self.metadata_mapping.items():
+            if field_name not in ["item_identifier", "source_system_identifier"]:
```
ghukill, line 135: Why would `item_identifier` or `source_system_identifier` be present in the mapping in the first place?

It seems as if the mapping would be constructed specifically for this application, and therefore could be omitted.

dsc/workflows/base/base_workflow.py
```diff
@@ -0,0 +1,199 @@
...
+
+                field_value = item_metadata.get(field_mapping["source_field_name"])
+                if field_value:
```
ghukill, line 138: I understand now that the mappings are complex objects and can have things like `delimiter` and `language`.

At a per workflow, per mapping level, is it worth considering adding an **optional** `required` property to the target field mapping?  Then you could do something like:

```python
if not field value and field_mapping.get("required",False):
  # raise an exception because an item didn't have a field that is required for this workflow/mapping
```

dsc/workflows/base/base_workflow.py
```diff
@@ -0,0 +1,199 @@
...
+            )
+            sqs_client.send_submission_message(
+                item_submission.item_identifier,
+                self.workflow_name,
+                self.output_queue,
+                self.submission_system,
+                self.collection_handle,
+                item_submission.metadata_uri,
+                item_submission.bitstream_uris,
+            )
```
ghukill, line 76: I'm happy to wait on this discussion if it would be helpful to tease out when this is actually built out.

Given this proposal, it feels like it could be a little difficult to test with this approach, where both `upload_space_metadata()` and `send_submission_message()` appear to be crafting objects _before_ they send them off into the ether.

What about a more verbose, but potentially easier to test flow:

```python
for item_submission in self.item_submissions_iter():

    # prepare dspace metadata and upload to S3
    dspace_metadata = item_submission.prepare_dspace_metadata()
    self.upload_dspace_metadata(metadata_s3_key, dspace_metadata)
    
    # send SQS message to DSS input queue
    submission_sqs_message = item_submission.prepare_sqs_input_message()
    sqs_client.publish_submission_message(submission_sqs_message)
```

This makes some assumptions that you may or may not agree with:
- the `BaseWorkflow` class would have the functionality to perform uploads to S3
- the `ItemSubmission` class could only prepare _payloads_ that will get sent
  - this removes the need for `ItemSubmission` to know anything about where they will go (S3 buckets, queues, 
etc.; all of which are sort of set on the workflow instance itself)
- the `ItemSubmission` class is what prepares SQS messages, but the `SQSClient` actually sends it
  - or, perhaps the `BaseWorkflow` could have a method like `submit_to_dss_input_queue()` which itself would utilize the `SQSClient` for actually publishing the message, but it would still be the _workflow_ that kicked that off and was ultimately responsible for it

Just food for thought.  But I realize this just a stub, and will get more attention when that part is written.

ghukill (commented):
Updates are looking good!  The call earlier today was helpful in understanding some of the changes.

Left a few comments.

dsc/workflows/base/base_workflow.py
ehanson8: You're correct, currently everything is DSpace@MIT but eventually I would like DSS to point to Dome for DDC uploads so this is to accommodate that. And yes, defining the init params on the derived workflow classes makes a lot of sense since they would largely be static for the workflow. `s3_prefix` and `collection_handle` the only ones that could potentially be dynamic, though I'd like to lock that down with the content owners where possible.

dsc/workflows/base/base_workflow.py
ehanson8: `metadata_s3_key` is better, great suggestion!

dsc/workflows/base/base_workflow.py
ehanson8: Good call, thanks!

dsc/workflows/base/base_workflow.py
ehanson8: I am fairly confident that it will!

dsc/workflows/base/base_workflow.py
ehanson8: `item_identifier` is provided by stakeholders in the CSV but may require additional parsing in the `get_item_identifier` method beyond retrieving it from the CSV (e.g. could be the full filename when we want to omit the extension). 

`source_system_identifier` is for creating reports used to update related metadata in other repositories (e.g. a digital object URI ArchivesSpace is the big use case) with the newly published DSpace handle. So it needs to be there for linking but isn't a part of the DSpace metadata that's generated. My expectation is this will be done in the `finalize` command. But second thought, it might be better to omit `source_system_identifier` until I'm ready to fully implement that functionality in the `ArchivesSpace` workflow, thoughts?

dsc/workflows/base/base_workflow.py
ehanson8: Good call! `title` is the only required DSpace field but certain workflows may need others, I'll implement it for title at this point

dsc/workflows/base/base_workflow.py
ehanson8: This is very helpful, things are already shifting from this proposed code and testing was a major consideration. This will be ready for full review in the next PR after I refactor the SQS functionality!

dsc/workflows/base/base_workflow.py
ehanson8: I like this and am currently implementing!

dsc/workflows/base/base_workflow.py
ghukill: > But second thought, it might be better to omit source_system_identifier until I'm ready to fully implement that functionality in the ArchivesSpace workflow, thoughts?

Might not be a bad idea?  We've been trying a similar approach -- and it's hard -- in the parquet work, to keep the early passes really tight and focused.  I think it's okay to stub something when it's almost gauranteed to show up later, and perhaps the contours of seeing it helps understand the current code to review.  But yeah, maybe it'll be easier to reason about, name, and organize when things settle into place a bit more.  Would make sense to me to wait.

ehanson8: Removed!

tests/test_base_workflow.py
```diff
@@ -0,0 +1,49 @@
...
+    base_workflow_instance,
+    item_metadata,
+):
+    assert base_workflow_instance.create_dspace_metadata(item_metadata) == {
+        "metadata": [
+            {"key": "dc.title", "language": "en_US", "value": "Title"},
+            {"key": "dc.contributor", "language": None, "value": "Author 1"},
+            {"key": "dc.contributor", "language": None, "value": "Author 2"},
+        ]
+    }
```
jonavellecuerdo, line 35: Hmm, I'm curious how you determine whether to add "_success" or "raises_exception" / "raise_error" to the test name? 🤔 

I brought this up with @ghukill while we were working on tests in `timdex-dataset-api`. I think some consistency or setting a standard per project would be nice! For instance...

* Given that [this test](https://github.com/MITLibraries/dspace-submission-composer/pull/21/files#diff-4aef579184063b7807d85afdb351da1571f6b608b597e746a3045eca07437f1aR4) is a single method with a single test, I'd expect [this test](https://github.com/MITLibraries/dspace-submission-composer/pull/21/files#diff-f8a31e340d9b850865bff75a6a3f13de37ab5cdcfc67ea2a2f9086332ea880d3R25) and [this other test](https://github.com/MITLibraries/dspace-submission-composer/pull/21/files#diff-f8a31e340d9b850865bff75a6a3f13de37ab5cdcfc67ea2a2f9086332ea880d3R25) should also include "_success".

ehanson8: You are correct, good catch!

dsc/workflows/base/__init__.py
```diff
@@ -0,0 +1,207 @@
...
+        """Generate a batch of item submissions for the DSpace Submission Service.
+
+        MUST NOT be overridden by workflow subclasses.
+        """
+        for item_metadata in self.batch_metadata_iter():
+            item_identifier = self.get_item_identifier(item_metadata)
+            logger.info(f"Processing submission for '{item_identifier}'")
+            metadata_s3_key = f"{self.s3_prefix}/{item_identifier}_metadata.json"
+            dspace_metadata = self.create_dspace_metadata(item_metadata)
+            self.validate_dspace_metadata(dspace_metadata)
```
ghukill, line 93: Looking at this now, I could envision validating the generated metadata _inside_ the creation.  But I think there are arguably advantages to doing it this way as well.  So NOT a request, just an observation, in case the pattern emerges elsewhere.

dsc/item_submission.py
```diff
@@ -0,0 +1,30 @@
...
+    """A class to store the required values for a DSpace submission."""
+
+    dspace_metadata: dict[str, Any]
+    bitstream_uris: list[str]
+    metadata_s3_key: str
+    metadata_uri: str = ""
```
ghukill, line 18: Minor, but maybe worth considering renaming `metadata_s3_uri` to match the pattern of `metadata_s3_key`.

dsc/item_submission.py
ghukill, line 1: Taking some of the other file reorganizing and renaming into consideration -- and thanks BTW, it's feeling good to navigate around! -- I could envision this file called something more high level like `items.py`.  _If_ other classes ever made sense to add specific to items, it would be a natural place for it.

I don't have evidence or even a pointable philosophy to support it, but when the classname mirrors the filename, it feels a bit off.  While we may not need more "item" type classes, it's tight coupling between file and class names.

Totally optional.

ghukill (approved):
All said, I feel comfortable approving!  Truly nice work all around, and thanks for the discussions.  It's one of the harder parts of a new project; establishing some patterns when parts are still stubbed or unfinished.

It's feeling opinionated enough that I can envision where future functionality may land, what it'll be called, etc.  But, also simple and flexible enough that if future functionality requires some shifting, that would be doable as well.

Left a couple of comments, but wholly optional.  Approved as-is, with or without changes.

dsc/item_submission.py
ehanson8: Updating to `metadata_s3_uri` and `bitstream_s3_uris` for clarity and consistency

jonavellecuerdo (approved):

PR #49, opened 2025-01-01
IN 1095 SimpleCSV workflow

jonavellecuerdo:

Implement the 'base' workflow class `SimpleCSV`. This SimpleCSV workflow is derived from the `BaseWorkflow` class and is an abstract class itself; the class is not meant to be called directly and will be used to derived stakeholder-specific workflow modules.

For now, reviewing the added unit tests should be sufficient for purposes of reviewing. 

YES - `smart_open` added to **packages**

YES - In my discussion with @ehanson8 , we proposed the following folder structure on S3 for `SimpleCSV`-dependent workflows.

**Note:** The "batch-XX" folders are arbitrary names assigned to "batches" of bitstreams uploaded by stakeholders.

```text
└── dsc/
    ├── imaging_lab/
    │   ├── batch-01/
    │   │   ├── metadata.csv
    │   │   └── <bitstream_files>
    │   └── archived/
    │       └── batch-00/
    │           ├── metadata.csv
    │           └── <bitstream_files>
    └── sccs/
        ├── batch-01/
        │   ├── metadata.csv
        │   └── <bitstream_files>
        └── archived/
            └── batch-00/
                ├── metadata.csv
                └── <bitstream_files>
```

- https://mitlibraries.atlassian.net/browse/IN-1095

### Developer

### Code Reviewer(s)

### DIFF
--- a/dsc/utilities/aws/s3.py
+++ b/dsc/utilities/aws/s3.py
@@ -58,23 +58,51 @@ class S3Client:
         return response
 
     def get_files_iter(
-        self, bucket: str, file_type: str, excluded_key_prefix: str
+        self,
+        bucket: str,
+        prefix: str = "",
+        file_identifier: str = "",
+        file_type: str = "",
+        exclude_prefixes: list[str] | None = None,
     ) -> Iterator[str]:
-        """Retrieve file based on file type, bucket, and without excluded prefix.
+        """Yield object keys for files stored on S3.
+
+        Results can be filtered by prefix, file identifier, and file type.
+        This method can also exclude files containing certain prefixes
+        (i.e., to skip files in 'archived/' folder).
 
         Args:
-            bucket: The S3 bucket to search.
-            file_type: The file type to retrieve.
-            excluded_key_prefix: Files with this key prefix will not be retrieved.
+            bucket (str): S3 bucket name.
+            prefix (str): Filter file list by prefix (i.e., subfolder in S3 bucket).
+            file_identifier (str, optional): Filter file list by unique identifier
+                in filename. Defaults to an empty string ("").
+            file_type (str, optional): Filter file list by file type (extension).
+                Defaults to an empty string ("").
+            exclude_prefixes (list[str], optional): Filter file list by excluding
+                file keys containing certain prefixes (e.g., 'archived').
+
+        Yields:
+            Iterator[str]: Files matching filters.
         """
         paginator = self.client.get_paginator("list_objects_v2")
-        page_iterator = paginator.paginate(Bucket=bucket)
+        page_iterator = paginator.paginate(Bucket=bucket, Prefix=prefix)
+
+        if exclude_prefixes is None:
+            exclude_prefixes = []
 
         for page in page_iterator:
-            files = [
-                content["Key"]
-                for content in page["Contents"]
-                if content["Key"].endswith(file_type)
-                and excluded_key_prefix not in content["Key"]
-            ]
-            yield from files
+            if page.get("Contents") is None:
+                return
+
+            for content in page["Contents"]:
+                if (
+                    content["Key"].endswith(file_type)
+                    and file_identifier in content["Key"]
+                ):
+                    if any(
+                        exclude_prefix in content["Key"]
+                        for exclude_prefix in exclude_prefixes
+                    ):
+                        continue
+
+                    yield content["Key"]
--- a/dsc/workflows/base/__init__.py
+++ b/dsc/workflows/base/__init__.py
@@ -26,7 +26,7 @@ class BaseWorkflow(ABC):
         email_recipients: list[str],
         metadata_mapping: dict,
         s3_bucket: str,
-        s3_prefix: str | None,
+        s3_prefix: str,
         collection_handle: str,
         output_queue: str,
     ) -> None:
@@ -53,7 +53,7 @@ class BaseWorkflow(ABC):
         self.email_recipients: list[str] = email_recipients
         self.metadata_mapping: dict = metadata_mapping
         self.s3_bucket: str = s3_bucket
-        self.s3_prefix: str | None = s3_prefix
+        self.s3_prefix: str = s3_prefix
         self.collection_handle: str = collection_handle
         self.output_queue: str = output_queue
 
@@ -81,11 +81,11 @@ class BaseWorkflow(ABC):
 
     @final
     def item_submissions_iter(self) -> Iterator[ItemSubmission]:
-        """Generate a batch of item submissions for the DSpace Submission Service.
+        """Yield item submissions for the DSpace Submission Service.
 
         MUST NOT be overridden by workflow subclasses.
         """
-        for item_metadata in self.batch_metadata_iter():
+        for item_metadata in self.item_metadata_iter():
             item_identifier = self.get_item_identifier(item_metadata)
             logger.info(f"Processing submission for '{item_identifier}'")
             metadata_s3_key = f"{self.s3_prefix}/{item_identifier}_metadata.json"
@@ -99,7 +99,7 @@ class BaseWorkflow(ABC):
             yield item_submission
 
     @abstractmethod
-    def batch_metadata_iter(self) -> Iterator[dict[str, Any]]:
+    def item_metadata_iter(self) -> Iterator[dict[str, Any]]:
         """Iterate through batch metadata to yield item metadata.
 
         MUST be overridden by workflow subclasses.
--- a/dsc/workflows/base/simple_csv.py
+++ b/dsc/workflows/base/simple_csv.py
@@ -1,0 +1,67 @@
+import csv
+import logging
+from typing import Any, Iterator
+
+import smart_open
+
+from dsc.workflows.base import BaseWorkflow
+from dsc.utilities.aws import S3Client
+
+logger = logging.getLogger(__name__)
+
+
+class SimpleCSV(BaseWorkflow):
+    def batch_metadata_iter(
+        self, metadata_file: str = "metadata.csv"
+    ) -> Iterator[dict[str, Any]]:
+        """Yield dicts of item metadata from metadata CSV file.
+
+        Args:
+            metadata_file (str, optional): Metadata CSV filename with
+                the filename extension (.csv) included.
+                Defaults to 'metadata.csv'.
+
+        Yields:
+            Iterator[dict[str, Any]]: Item metadata.
+        """
+        with smart_open.open(
+            f"s3://{self.s3_bucket}/{self.s3_prefix}/{metadata_file}"
+        ) as csvfile:
+            reader = csv.DictReader(csvfile)
+            for metadata_entry in reader:
+                yield metadata_entry
+
+    def get_bitstream_uris(self, item_identifier: str) -> list[str]:
+        """Get S3 URIs for bitstreams for a given item.
+
+        This method uses S3Client.get_files_iter to get a list of files
+        on S3 stored at s3://bucket/prefix/ and includes the 'item_identifier'
+        in the object key.
+
+        - If the exact filename is provided as 'item_identifier', only
+          one bitstream is retrieved.
+        - If a prefix is provided as 'item_identifier', one or more
+          bitstreams are retrieved.
+
+        Args:
+            item_identifier (str): Item identifier used to filter bitstreams.
+
+        Returns:
+            list[str]: Bitstream URIs for a given item.
+        """
+        s3_client = S3Client()
+        return [
+            bitstream_uri
+            for bitstream_uri in s3_client.get_files_iter(
+                bucket=self.s3_bucket,
+                prefix=self.s3_prefix,
+                file_identifier=item_identifier,
+            )
+        ]
+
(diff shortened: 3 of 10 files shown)


dsc/workflows/base/simple_csv.py
```diff
@@ -0,0 +1,66 @@
...
+          one bitstream is retrieved.
+        - If a prefix is provided as 'item_identifier', one or more
+          bitstreams are retrieved.
+
+        Args:
+            item_identifier (str): Item identifier used to filter bitstreams.
+
+        Returns:
+            list[str]: Bitstream URIs for a given item.
+        """
```
jonavellecuerdo, line 50: I was interested in getting both of your thoughts around the retrieval of bitstreams using the `item_identifier` from the metadata CSV file. 🤔 

@ehanson8 and I discussed what values were typically provided for this column by stakeholders, and we arrived at the following logic: 
> The item identifier can be the exact filename or just the prefix.
> - If the exact filename is provided, only one bitstream will be retrieved.
> - If the prefix is provided, one or more bitstreams will be retrieved. 

To clarify: 
1. **Full filename is provided:** `"abc123.pdf"`
2. **Prefix (of item identifier) is provided:** `"abc123"`

Should `SimpleCSV.get_bitstream_uris` raise an error if `>1` bitstreams are retrieved when `item_identifier` is a filename?
For instance, what if there were two bitstreams:

```
dsaps/stakeholder/batch-00/file1-abc123.pdf
dsaps/stakeholder/batch-00/file2-abc123.pdf
```

It seems that it _should_ capture both files, but the logic we arrived at then breaks down. 🤔

ehanson8: That example doesn't reflect how the `item_identifier` appears, it is always a prefix so it would be like this: 

```
dsaps/stakeholder/batch-00/abc123-file1.pdf
dsaps/stakeholder/batch-00/abc123-file2.pdf
```

Which would not be a problem at all, we would just get 2 bitstreams for that `item_identifier` because we're using `.startswith(item_identifier)` not `item_identifier in` (until we get to use cases where we need regex like the Archivematica output). Happy to huddle and discuss further!

ghukill: Perhaps relevant to this discussion, I keep coming back to [`prefix` and `file_identifier` in the `S3Client`](https://github.com/MITLibraries/dspace-submission-composer/pull/49/files#diff-051f9b929af8902d0b16bbb6966ea916f9f1923fe4828796f3a3f51b407024c6R76-R78).  

Wouldn't any given `file_identifier` -- which I keep thinking is synonmous with the filename minus the extension -- be part of an S3 prefix as well?

For example, I get the same results from the local Minio test example with this tweak for `S3Client.get_bitstream_uris()`:

```python
return list(
    # s3_client.get_files_iter(
    #     bucket=self.s3_bucket,
    #     prefix=self.s3_prefix,
    #     file_identifier=item_identifier,
    # )
    s3_client.get_files_iter(
        bucket=self.s3_bucket,
        prefix="/".join([self.s3_prefix, item_identifier]),
    )
)
```

Not to imply that the `"/".join(...)` is a good route to go, just that the `file_identifier` feels more like an extension of the S3 key prefix.  And, perhaps this touches on @ehanson8's point about possible regex for Archivematica.

ehanson8: > Wouldn't any given `file_identifier` -- which I keep thinking is synonmous with the filename minus the extension -- be part of an S3 prefix as well?

Yes, it would be a part of the prefix but it is not always synonymous with filename minus the extension as there may be `_001` or `_a` as well. Let's have a huddle when it's convenient for everyone, I think we can clarify `file_identifier` vs `item_identifier`, I agree it is murky now that we're trying to accommodate multiple workflows

jonavellecuerdo: The three of us had a discussion via Slack huddle and led to proposal of additional updates, which are implemented via commit https://github.com/MITLibraries/dspace-submission-composer/pull/49/commits/99ae5a1c3c492384013fa1210d24d3da3f74a501.

dsc/workflows/base/simple_csv.py
```diff
@@ -0,0 +1,67 @@
...
+
+import smart_open
+
+from dsc.utilities.aws import S3Client
+from dsc.workflows.base import BaseWorkflow
+
+logger = logging.getLogger(__name__)
+
+
+class SimpleCSV(BaseWorkflow):
```
ghukill, line 14: Not required, but I think a docstring for this class could be helpful.  Maybe short-and-sweet for the time being, in case there are learnings for actual workflows that extend this, that would be helpful to work back into this base class.

But this feels like a pretty foundational class in this application, and even just _stubbing_ the docstring section feels like it could be helpful.

tests/conftest.py
ghukill, line 1: Commenting at the file level...

I think we may need to handle `AWS_ENDPOINT_URL` for the `_test_env` fixture.  If I have this set in my `.env` file -- as I did for Minio testing -- the tests fails as they pickup this altered endpoint URL.

Could even be as simple as _unsetting_ it:
```python
monkeypatch.delenv("AWS_ENDPOINT_URL")
```

tests/conftest.py
```diff
@@ -28,30 +30,44 @@ def _test_env(monkeypatch):
...
             return item_metadata["item_identifier"]
 
         def get_bitstream_uris(self, item_identifier):
             bitstreams = [
-                "s3://dsc/workflow/folder/123_01.pdf",
-                "s3://dsc/workflow/folder/123_02.pdf",
-                "s3://dsc/workflow/folder/456_01.pdf",
+                "s3://dsc/base/batch-aaa/123_01.pdf",
+                "s3://dsc/base/batch-aaa/123_02.pdf",
+                "s3://dsc/base/batch-aaa/456_01.pdf",
```
ghukill, line 43: I think seeing the simulated filepaths in S3 here is nice!  Can imagine how this will work well over time.

ghukill (changes requested):
Overall, looks good to me and would have approved.  

Requesting changes for a small update to `conftest.py` to handle if `AWS_ENDPOINT_URL` is set in a user's `.env` file.

dsc/workflows/base/simple_csv.py
ehanson8: Agreed!

tests/conftest.py
ehanson8: Same happened to me, agree with this suggestion!

dsc/utilities/aws/s3.py
```diff
@@ -58,23 +58,51 @@ def put_file(
         return response
 
     def get_files_iter(
-        self, bucket: str, file_type: str, excluded_key_prefix: str
+        self,
+        bucket: str,
+        prefix: str = "",
+        file_identifier: str = "",
+        file_type: str = "",
+        exclude_prefixes: list[str] | None = None,
```
ehanson8, line 66: Could this just be `exclude_prefixes: list[str] = []`? You could remove lines 90-91:

```
if exclude_prefixes is None:
   exclude_prefixes = []
```

tests/test_s3.py
```diff
@@ -1,12 +1,14 @@
+import json
+import re
 from http import HTTPStatus
 
 import pytest
 from botocore.exceptions import ClientError
 
 
-def test_s3_archive_file_in_bucket(mocked_s3, s3_client):
+def test_s3_client_archive_file_with_new_key_success(mocked_s3, s3_client):
```
ehanson8, line 9: Good rename!

dsc/workflows/base/__init__.py
```diff
@@ -26,37 +26,44 @@ def __init__(
...
-            workflow.
-            s3_prefix: The S3 prefix used for objects in this workflow. This prefix does
-            NOT include the bucket name.
-            collection_handle: The handle of the DSpace collection to which submissions
-            will be uploaded
-            output_queue: The SQS output queue used for retrieving result messages from
-            the workflow's submissions.
+            workflow_name (str): The name of the workflow.
+            submission_system (str): The system to which item submissions will be sent
+                (e.g. DSpace@MIT).
```
ehanson8, line 38: A global comment, not sure that the type hints are needed in the docstrings since they appear in the signature

tests/test_s3.py
```diff
@@ -15,53 +17,103 @@ def test_s3_archive_file_in_bucket(mocked_s3, s3_client):
...
+        archived_key_prefix="workflow/folder/archived",
     )
+
+    assert list(
+        s3_client.files_iter(
+            bucket="dsc",
+            prefix="workflow/folder",
+            exclude_prefixes=["workflow/folder/archived"],
+        )
+    ) == ["workflow/folder/aaa.pdf"]
```
ehanson8, line 119: Get set of tests for `files_iter`!

tests/test_simple_csv_workflow.py
```diff
@@ -0,0 +1,44 @@
...
+    metadata_iter = simple_csv_workflow_instance.item_metadata_iter(
+        metadata_file="metadata.csv"
+    )
+    assert next(metadata_iter) == item_metadata
+
+
+def test_simple_csv_workflow_get_item_identifier_success(
+    simple_csv_workflow_instance, item_metadata
+):
+    assert simple_csv_workflow_instance.get_item_identifier(item_metadata)
```
ehanson8, line 16: Could we `assert` the actual value retrieved by `get_item_identifier`?

ehanson8 (commented):
Looking good! Just a few suggestions

dsc/workflows/base/__init__.py
ghukill: I think I agree.   And, I think it's okay to _informally_ mention the type in the explanation of the argument if it helps reason about how to use.

tests/conftest.py
jonavellecuerdo: Ah, it failed for me as well, so this is a good suggestion! I will work on this.

tests/test_s3.py
jonavellecuerdo: I've been trying to keep up with my test naming conventions / structure since our work on the Transmog refactor! 🤓

dsc/workflows/base/__init__.py
jonavellecuerdo: Ah, I see! Whenever I revise a docstring, I tend to let the [autoDocstring extension for VSCode](https://marketplace.visualstudio.com/items?itemName=njpwerner.autodocstring) automatically generate the docstring and I have it set to the [Google Docstring format](https://github.com/NilsJPWerner/autoDocstring/blob/HEAD/docs/google.md), which automatically includes parameter types.

Hmm...do I remove all the typehints from all classes and methods (that I most likely introduced 😅 ) now or ...? I did look at the autoDocstring docs and it seems like I can define a custom template, but I'd like to revisit that process later.

Would love some additional input on how to proceed!

ehanson8: Modifying the template sounds good for the future and I would remove them. I think find/replace on something like `(str):` would get a lot of them

dsc/utilities/aws/s3.py
ehanson8: Ah, did not know that, thank you for the heads up!

tests/test_s3.py
jonavellecuerdo: Thank you! Minor, but I made a small fix to follow the format:

```text
<workflow>/batch-aaa/<123>.pdf
```

when mocking S3 prefixes for consistency!

dsc/workflows/base/__init__.py
ghukill: I'm not philosophically opposed to the docstrings having the types, but they would be my first bet to fall out of sync with the code.  I recall in ABDiff when we did a lot of signature rewriting, that the lengthy docstrings were stale pretty quick.  That's one thing I really like about actual type hints in code, is they _must_ be up-to-date and correct (well, mostly).

tests/test_simple_csv_workflow.py
jonavellecuerdo: Change has been made!

tests/conftest.py
jonavellecuerdo: Thank you again for pointing this out! Updated.

dsc/workflows/base/simple_csv.py
jonavellecuerdo: Short docstring added!

tests/conftest.py
```diff
@@ -22,36 +24,51 @@ def _test_env(monkeypatch):
     monkeypatch.setenv("AWS_ACCESS_KEY_ID", "testing")
     monkeypatch.setenv("AWS_SECRET_ACCESS_KEY", "testing")
     monkeypatch.setenv("DSS_INPUT_QUEUE", "mock-input-queue")
+    monkeypatch.delenv("AWS_ENDPOINT_URL", raising=False)
```
ghukill, line 27: 👍

ghukill (approved):

dsc/workflows/base/__init__.py
jonavellecuerdo: I removed the docstrings but kept the indentation for multiline parameter definitions since I think it helps with readability!

ehanson8 (approved):
Great work!

PR #56, opened 2025-01-06
In 1103 reconcile cli command

ehanson8:
Add workflow functionality in addition to a `reconcile` CLI command. Reviewing each commits separately would be ideal.

The `reconcile` command would be run manually before each DSC submission batch to ensure that the metadata and bitstreams are properly aligned, ensuring there no bitstreams without metadata and no metadata without at least one bitstream. After that is confirmed, the soon-to-be-written `deposit` CLI command would be run for the batch.

Not possible yet

YES

NO

- https://mitlibraries.atlassian.net/browse/IN-1103

### Developer

### Code Reviewer(s)


dsc/cli.py
```diff
@@ -6,27 +6,117 @@
...
+    )
+
+
+@main.command()
+@click.pass_context
+@click.option(
+    "-f",
+    "--file_type",
+    help="Optional parameter to filter bitstreams to specified file type",
+)
```
ghukill, line 87: Can you expand on when and why we might need to filter bistreams based on file type?

ehanson8: On a few occasions, DDC had `tiff` (archival and huge) and `jpg` (access and small) files with identical identifiers in the same directory. Though I should probably check with them if that will still be the case, given that we're using the Archivematica output now. And given that we're asking stuff to be uploaded to S3, filtering to the correct file types should probably happen as a part of the upload process for other stakeholders

ghukill: Thanks for explaining!  I have a hunch that this may relate to some still forming questions I had about the reconciling.

Thought experiment: what if instead of "loose" files in the batch folder in S3, where you're planning on matching files per an identifier prefix, each item had it's own folder within the batch folder?  So if I have a derived item identifier `foo-bar-123` in my CSV file, I'd expect a folder `foo-bar-123` and _every_ file under that folder was to be included as a bitstream for that item.  Feels like this could be easier _or_ harder for stakeholders; easier to reason at a glance about files being present for an item, but perhaps a bit more fillle management.

As it relates ot the reconciling, then you could log a warning or raise an error (unsure which you have planned) if there is an item identifier _without_ a folder of files that will become bitstreams.   Or vice versa, a folder of files that don't have an item in the metadata.

Don't mean to suggest a big change, maybe more of a thought experiment to help me understand the logic in the reconciling CLI command.  And maybe there are considerations for non-CSV workflows where this breaks down.

dsc/config.py
```diff
@@ -5,6 +5,12 @@
 
 import sentry_sdk
 
+WORKFLOWS = {
+    "test": {
+        "workflow-path": "tests.conftest.TestBaseWorkflow",
```
ghukill, line 10: If you are going to go the route of a dictionary/JSON object to define what workflows are called, and where to find them, I might suggest changing `workflow-path` to something like `workflow_classpath` which I think explicitly indicates this is the import pattern for the workflow class itself.

dsc/workflows/base/__init__.py
```diff
@@ -21,51 +24,73 @@
...
+        """
+        workflow_class = cls.get_workflow(workflow_name)
+        return workflow_class(
+            collection_handle=collection_handle,
+            batch_id=batch_id,
+        )
+
+    @final
+    @classmethod
+    def get_workflow(cls, workflow_name: str) -> type[BaseWorkflow]:
```
ghukill, line 83: I'm curious if you considered the `__subclasses()__` approach for discovering workflow classes?  

I'm not opposed to the dictionary/JSON object configuration approach either, it's certainly worked in other projects (e.g. Transmog).  But in those situations, there was often more configuration about the thing in that config object.  As-is, I think there is a bit of a tension between the class level attributes and this config object.  How do you decide what goes where?  

If the config object is _only_ for discovering the python classes based on a name, I do think the `__subclasses()__` approach might be worth considering.

ghukill (commented):
I'm understanding how this application is taking shape, and continue to think the overall pattern is a great one (i.e. workflows).

Leaving a "Comment" review at this time with somewhat high level questions.

dsc/cli.py
```diff
@@ -6,27 +6,117 @@
...
+def reconcile(ctx: click.Context, file_type: str) -> None:
```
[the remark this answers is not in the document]
ehanson8: I was having a similar thought yesterday evening, I agree with you that it should move to the the `BaseWorkflow` and then could be overridden by any derived workflow classes. Great suggestion!

dsc/cli.py
```diff
@@ -6,27 +6,117 @@
...
+    "--batch_id",
+    help="The S3 prefix for the batch of DSpace submissions",
+    required=True,
+)
+@click.option(
+    "-c",
+    "--collection_handle",
+    help="The handle of the DSpace collection to which the batch will be submitted",
+    required=True,
+)
```
jonavellecuerdo, line 39: Can we use dashes to separate multi-word command line arguments to follow Unix/Linux conventions? 🤔

dsc/cli.py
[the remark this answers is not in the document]
jonavellecuerdo: I think I like this idea! Given that `reconcile` is the first step in the DSC workflow for _any_ workflow, the `BaseWorkflow` module should house this complex logic!

jonavellecuerdo (changes requested):
@ehanson8 I think this is a great start! As I reviewed the updates in this PR, it made me wonder about environment variables and wondered if we could have a team discussion to discuss the workflow implementation in more detail?

dsc/cli.py
ehanson8: Good call, updated!

dsc/workflows/base/__init__.py
ehanson8: Found a much cleaner approach with `__subclasses__`, great suggestion!

dsc/workflows/base/__init__.py
```diff
@@ -21,51 +28,101 @@
...
+            if workflow_name == workflow_class.workflow_name:
+                return workflow_class
+        raise InvalidWorkflowNameError(f"Invalid workflow name: {workflow_name} ")
+
+    def reconcile_bitstreams_and_metadata(self) -> tuple[set[str], set[str]]:
+        """Reconcile bitstreams against metadata.
+
+        Generate a list of bitstreams without item identifiers and item identifiers
+        without bitstreams. Any discrepancies will be addressed by the engineer and
+        stakeholders as necessary.
```
ghukill, line 104: I really like this docstring.  I think too often we (at least I) get wrapped up in docstrings being about a technical explanation of technical code, I think this does a fantastic job of situating what this code _does_ in the context when when it's called and by whom.  Nice work, that's not an easy task.

ehanson8: Thanks, I'm getting more sensitive to that on this app as I realize the "why" for some of this is not self-evident if you haven't been doing this work in the past

dsc/workflows/base/__init__.py
```diff
@@ -21,51 +28,101 @@
...
+            for item_metadata in self.item_metadata_iter()
+        ]
+
+        # reconcile item identifiers against bitstreams
+        item_identifier_matches = match_item_identifiers_to_bitstreams(
+            bitstream_dict.keys(), item_identifiers
+        )
+        file_matches = match_bitstreams_to_item_identifiers(
+            bitstream_dict.keys(), item_identifiers
+        )
```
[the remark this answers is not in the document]
ehanson8: Good point on making them private methods, `build_bitstream_dict` can get both params from `self`. But even though the other 2 are just list comprehensions, they are involved enough that I personally think it reads and tests easier if they're still functions/methods

dsc/workflows/base/__init__.py
```diff
@@ -21,51 +24,73 @@
 class BaseWorkflow(ABC):
     """A base workflow class from which other workflow classes are derived."""
 
+    workflow_name: str = "base"
+    submission_system: str = "DSpace@MIT"
+    email_recipients: tuple[str] = ("None",)
+    metadata_mapping_path: str = ""
+    s3_bucket: str = ""
+    output_queue: str = ""
```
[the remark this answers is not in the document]
jonavellecuerdo: As noted by @ehanson8 in Slack, the questions/comments above do not need to be addressed via this PR, but it has prompted a discussion about workflow implementation to be addressed in future PRs!

dsc/workflows/base/__init__.py
```diff
@@ -21,51 +26,152 @@
...
+        item_identifier_matches = self._match_item_identifiers_to_bitstreams(
+            bitstream_dict.keys(), item_identifiers
+        )
+        file_matches = self._match_bitstreams_to_item_identifiers(
+            bitstream_dict.keys(), item_identifiers
+        )
+        logger.info(f"Item identifiers and bitstreams matched: {item_identifier_matches}")
+        no_bitstreams = set(item_identifiers) - set(item_identifier_matches)
+        no_item_identifiers = set(bitstream_dict.keys()) - set(file_matches)
+        return no_bitstreams, no_item_identifiers
```
jonavellecuerdo, line 122: Hmm, two questions: 

1. What do you think of the proposed name changes below? A little wordy but I think it helps with clarity (at least for someone like me who struggles with logic that involves matching 😅).
```python
# reconcile item identifiers against bitstreams
item_identifiers_with_bitstream_matches = self._match_item_identifiers_to_bitstreams(
    bitstream_dict.keys(), item_identifiers
)
bitstreams_with_item_identifier_matches = self._match_bitstreams_to_item_identifiers(
    bitstream_dict.keys(), item_identifiers
)

logger.info(f"Item identifiers and bitstreams matched: {item_identifier_matches}")
no_bitstreams = set(item_identifiers) - set(item_identifier_with_bitstream_matches)
no_item_identifiers = set(bitstream_dict.keys()) - set(bitstreams_with_item_identifier_matches)
return no_bitstreams, no_item_identifiers
```

2. Can we clarify the log message in line 119? These are technically:
> Item identifiers from batch metadata with matching bitstreams: <item identifiers>

jonavellecuerdo (changes requested):
@ehanson8 Changes are looking good! Just have two comments (potential change requests).

dsc/workflows/base/__init__.py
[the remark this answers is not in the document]
jonavellecuerdo: We synced up on env vars and arrived at the following decision:

* Explicitly set class attributes in the workflow module: 
```python
class BaseWorkflow(ABC):
    """A base workflow class from which other workflow classes are derived."""

    workflow_name: str = "base"
    submission_system: str = "DSpace@MIT"
    email_recipients: tuple[str] = ("None",)
    metadata_mapping_path: str = ""
    s3_bucket: str = "dsc"
    output_queue: str = "dsc-unhandled"
```
* Revisit use of env vars _later_!

dsc/workflows/base/__init__.py
jonavellecuerdo: Ooh, I think this needs a small fix (and maybe slightly simpler naming convention): 

```python
# reconcile item identifiers against bitstreams
item_identifiers_with_bitstreams = (
    self._match_item_identifiers_to_bitstreams(
        bitstream_dict.keys(), item_identifiers
    )
)

bitstreams_with_item_identifiers = self._match_bitstreams_to_item_identifiers(
    bitstream_dict.keys(), item_identifiers
)

logger.info(
    "Item identifiers from batch metadata with matching bitstreams: "
    f"{item_identifiers_with_bitstreams}"
)

item_identifiers_without_bitstreams =  set(item_identifiers) - set(
    item_identifiers_with_bitstreams
)
bitstreams_without_item_identifiers = set(bitstream_dict.keys()) - set(bitstreams_with_item_identifiers)

return item_identifiers_without_bitstreams, bitstreams_without_item_identifiers
```
Essentially, my initial proposal was to rename `file_matches` variable as well!

---

This will align with the [log messages in the cli](https://github.com/MITLibraries/dspace-submission-composer/blob/IN-1103-reconcile-cli-command/dsc/cli.py#L81-L88).
* I'd suggest maybe updating the variable names in the `reconcile` CLI command as well!

Sorry for the confusion!

ehanson8: My bad, fixed!

jonavellecuerdo (approved):

ghukill (approved):
Really appreciate all the thoughtful discussions @ehanson8.  As discussed, further work will likely fill out pieces that feel a bit fuzzy, but this feels like a good foundation to begin with to build out workflows.

PR #68, opened 2025-01-10
In 1104 deposit command

ehanson8:
Refactor CLI functionality in addition adding a `deposit` CLI command.

Not possible yet

YES

NO

- https://mitlibraries.atlassian.net/browse/IN-1104

### Developer

### Code Reviewer(s)

### DIFF
--- a/dsc/cli.py
+++ b/dsc/cli.py
@@ -66,11 +66,11 @@ def main(
 ) -> None:
     ctx.ensure_object(dict)
     ctx.obj["start_time"] = perf_counter()
-    workflow = Workflow.load(
-        workflow_name=workflow_name,
+    workflow_class = Workflow.get_workflow(workflow_name)
+    workflow = workflow_class(
         collection_handle=collection_handle,
         batch_id=batch_id,
-        email_recipients=email_recipients,
+        email_recipients=tuple(email_recipients.split(",")),
         s3_bucket=s3_bucket,
         output_queue=output_queue,
     )
@@ -125,5 +125,4 @@ def deposit(ctx: click.Context) -> None:
     """Send a batch of item submissions to the DSpace Submission Service (DSS)."""
     workflow = ctx.obj["workflow"]
     logger.debug(f"Beginning submission of batch ID: {workflow.batch_id}")
-    submission_results = workflow.run()
-    logger.debug(f"Results of submission: {submission_results}")
+    workflow.run()
--- a/dsc/cli.py
+++ b/dsc/cli.py
@@ -6,7 +6,7 @@ from time import perf_counter
 import click
 
 from dsc.config import Config
-from dsc.workflows.base import BaseWorkflow
+from dsc.workflows.base import Workflow
 
 logger = logging.getLogger(__name__)
 CONFIG = Config()
@@ -29,8 +29,23 @@ CONFIG = Config()
 @click.option(
     "-b",
     "--batch-id",
-    help="The S3 prefix for the batch of DSpace submissions",
+    help="The S3 prefix for the batch of DSpace submission files",
     required=True,
+)
+@click.option(
+    "-s",
+    "--s3-bucket",
+    help="The S3 bucket containing the DSpace submission files",
+)
+@click.option(
+    "-o",
+    "--output-queue",
+    help="The SQS output queue for the DSS result messages",
+)
+@click.option(
+    "-e",
+    "--email-recipients",
+    help="The recipients of the submission results email",
 )
 @click.option(
     "-v", "--verbose", is_flag=True, help="Pass to log at debug level instead of info"
@@ -40,12 +55,22 @@ def main(
     workflow_name: str,
     collection_handle: str,
     batch_id: str,
+    s3_bucket: str | None,
+    output_queue: str | None,
+    email_recipients: tuple[str] | None,
     verbose: bool,  # noqa: FBT001
 ) -> None:
     ctx.ensure_object(dict)
     ctx.obj["start_time"] = perf_counter()
-
-    ctx.obj["workflow"] = BaseWorkflow.load(workflow_name, collection_handle, batch_id)
+    workflow = Workflow.load(
+        workflow_name=workflow_name,
+        collection_handle=collection_handle,
+        batch_id=batch_id,
+        s3_bucket=s3_bucket,
+        output_queue=output_queue,
+        email_recipients=email_recipients,
+    )
+    ctx.obj["workflow"] = workflow
 
     stream = StringIO()
     root_logger = logging.getLogger()
@@ -86,3 +111,15 @@ def reconcile(ctx: click.Context) -> None:
         logger.error(
             f"No item identifiers found for these bitstreams: {no_item_identifiers}"
         )
+    else:
+        logger.info("All item identifiers and bitstreams successfully matched")
+
+
+@main.command()
+@click.pass_context
+def deposit(ctx: click.Context) -> None:
+    """Send a batch of item submissions to the DSpace Submission Service (DSS)."""
+    workflow = ctx.obj["workflow"]
+    logger.debug(f"Beginning submission of batch ID: {workflow.batch_id}")
+    for response in workflow.run():
+        logger.debug(response)
--- a/dsc/cli.py
+++ b/dsc/cli.py
@@ -33,19 +33,23 @@ CONFIG = Config()
     required=True,
 )
 @click.option(
+    "-e",
+    "--email-recipients",
+    help="The recipients of the submission results email as a comma-delimited string, "
+    "this will override the workflow class's default value for this attribute",
+    required=True,
+)
+@click.option(
     "-s",
     "--s3-bucket",
-    help="The S3 bucket containing the DSpace submission files",
+    help="The S3 bucket containing the DSpace submission files, "
+    "this will override the workflow class's default value for this attribute",
 )
 @click.option(
     "-o",
     "--output-queue",
-    help="The SQS output queue for the DSS result messages",
-)
-@click.option(
-    "-e",
-    "--email-recipients",
-    help="The recipients of the submission results email",
+    help="The SQS output queue for the DSS result messages, "
+    "this will override the workflow class's default value for this attribute",
 )
 @click.option(
     "-v", "--verbose", is_flag=True, help="Pass to log at debug level instead of info"
@@ -55,9 +59,9 @@ def main(
     workflow_name: str,
     collection_handle: str,
     batch_id: str,
+    email_recipients: str,
     s3_bucket: str | None,
     output_queue: str | None,
-    email_recipients: tuple[str] | None,
     verbose: bool,  # noqa: FBT001
 ) -> None:
     ctx.ensure_object(dict)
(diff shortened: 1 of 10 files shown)


dsc/workflows/base/__init__.py
```diff
@@ -63,37 +69,45 @@ def metadata_mapping(self) -> dict:
...
         """
         workflow_class = cls.get_workflow(workflow_name)
-        return workflow_class(
-            collection_handle=collection_handle,
-            batch_id=batch_id,
-        )
+        filtered_kwargs = {
+            key: value for key, value in kwargs.items() if value is not None
+        }
+        return workflow_class(**filtered_kwargs)  # type: ignore [arg-type]
```
ehanson8, line 89: Not sure if I'm doing this right given the need for `# type: ignore [arg-type]` so let me know if there's a better way!

dsc/cli.py
```diff
@@ -29,9 +29,24 @@
...
+@click.option(
+    "-o",
+    "--output-queue",
+    help="The SQS output queue for the DSS result messages",
+)
+@click.option(
+    "-e",
+    "--email-recipients",
+    help="The recipients of the submission results email",
+)
```
ghukill, line 49: Feel free to disregard if not true, but could we update the docstring for these to indicate passing these will _override_ possible defaults in the Workflow definition itself?

dsc/cli.py
```diff
@@ -86,3 +111,15 @@ def reconcile(ctx: click.Context) -> None:
...
+
+
+@main.command()
+@click.pass_context
+def deposit(ctx: click.Context) -> None:
+    """Send a batch of item submissions to the DSpace Submission Service (DSS)."""
+    workflow = ctx.obj["workflow"]
+    logger.debug(f"Beginning submission of batch ID: {workflow.batch_id}")
+    for response in workflow.run():
+        logger.debug(response)
```
ghukill, line 125: This feels a little off for a root level CLI command behavior.  If I'm reading correctly, given only `DEBUG` logging, if logging were `INFO` you'd get no output from this CLI command?

A couple of thoughts/proposals.

First, I'd propose changing line 123 to `logger.info`.  That feels like a standard, helpful, baseline thing to log.

Second, a bit more complex, I might propose that `Workflow.run()` does not yield SQS responses, but instead aggregates them, possible runs some analysis, and returns a payload of some sort.  Apologies for not catching this on earlier review of `Workflow.run()`, just kind of noticed the implications here in the CLI command.

It looks as though `ItemSubmission.send_submission_message()` returns a "raw" `boto3` SQS message send receipt, which is good/fine.  I think there I may propose parsing that would be the `Workflow.run()` method.

What if it returned some kind of payload like a dictionary:

```python
{
    "success":True|False, # True if ALL successful, else False
    "submission_receipts":[list of receipts from ItemSubmission.send_submission_message()]
}
```

Unsure if there is anything else helpful that could be included here... something known after the SQS messages are successfully submitted.  

This way, `Workflow.run()` could return something substantial that exhausts all the underlying iterators, knowing it's basically only the base CLI command that would call it.

ghukill (changes requested):
I'm liking the direction, but a quick "Request Changes" before continuing with the review.  Left a comment about the return signature of `Workflow.run()` which I think has implications for the CLI command `deposit`.

dsc/cli.py
ehanson8: Good point on changing `logger.debug` to `logger.info` but the response from the DSS input is not terribly useful, an example below:
```
{'MD5OfMessageBody': 'f868d427c25f778ed0e003745ec3f7d7', 'MD5OfMessageAttributes': '1345a1b7906d650665ddeb7741ef4465', 'MessageId': '6f93ae95-cc59-4242-bf28-cca0c05019e5', 'ResponseMetadata': {'RequestId': 'AoCTow04FJuwTRqXqEs2LPSL0fVkmlQ1Zz71IGECqR8OcC9mVnuJ', 'HTTPStatusCode': 200, 'HTTPHeaders': {'server': 'amazon.com', 'date': 'Mon, 13 Jan 2025 11:46:55 GMT', 'x-amzn-requestid': 'AoCTow04FJuwTRqXqEs2LPSL0fVkmlQ1Zz71IGECqR8OcC9mVnuJ', 'x-amz-crc32': '3274493029'}, 'RetryAttempts': 0}}
```
I could do the success check you're suggesting based on the `HTTPStatusCode` and `MessageId` but not much beyond that

ghukill: Thanks for sharing the response.  I'd vote for:

1. ensure that all `HTTPStatusCode`'s are what's expected, and this would drive an overall success True/False
2. return a list of the `MessageId`; these can be helpful for debugging, as you can jump into AWS console and view those messages individually if needed

The overall suggestion is that `Workflow.run()` does not return an iterator, but a payload that communicates to the CLI if the "run" was successful.

ehanson8: A good approach, will do!

jonavellecuerdo (changes requested):

dsc/cli.py
jonavellecuerdo: Hmm but how is the `MessageId` derived and how (if it does) does it link back to the item identifier? 🤔

dsc/workflows/base/__init__.py
```diff
@@ -23,20 +23,20 @@
...
-    email_recipients: tuple[str] = ("None",)
     metadata_mapping_path: str = ""
-    s3_bucket: str = ""
-    output_queue: str = ""
 
     def __init__(
         self,
         collection_handle: str,
         batch_id: str,
+        email_recipients: tuple[str] = ("None",),
```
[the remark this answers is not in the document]
ehanson8: Good call, done!

dsc/cli.py
ehanson8: Pushed a [commit](https://github.com/MITLibraries/dspace-submission-composer/pull/68/commits/09e23ee9aad082092620801db3bdbc7ddf8740d7) with changes to address this

dsc/cli.py
```diff
@@ -121,5 +125,5 @@ def deposit(ctx: click.Context) -> None:
     """Send a batch of item submissions to the DSpace Submission Service (DSS)."""
     workflow = ctx.obj["workflow"]
     logger.debug(f"Beginning submission of batch ID: {workflow.batch_id}")
-    for response in workflow.run():
-        logger.debug(response)
+    submission_results = workflow.run()
+    logger.debug(f"Results of submission: {submission_results}")
```
ehanson8, line 129: Since I added a concise `logger.info` summary of the results inside `Workflow.run`, I think the full response with all of the item identifiers should be in `logger.debug` but let me know if you disagree!

dsc/workflows/base/__init__.py
```diff
@@ -184,17 +200,51 @@ def _match_item_identifiers_to_bitstreams(
...
+        """Run workflow to submit items to  the DSpace Submission Service.
+
+        Returns a dict with the attempted, successful, and failed submissions as well as
+        the results of each item submission organized by the item identifier.
+        """
+        submission_results: dict[str, Any] = {"success": True}
+        attempted_submissions = 0
+        successful_submissions = 0
+        failed_submissions = 0
         for item_submission in self.item_submissions_iter():
```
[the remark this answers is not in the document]
ghukill: Another approach could be to lean more directly into the items as the most useful artifact, and then derive stats from lists of successes and failures.   The advantage to this data-driven approach, in my opinion, is that it's easier to update or modify the stats logged or returned later; they are just math based on structured data.

Example:

```python
@final
def run(self) -> dict:
    """Run workflow to submit items to  the DSpace Submission Service.

    Returns a dict with the attempted, successful, and failed submissions as well as
    the results of each item submission organized by the item identifier.
    """

    items = {"succeeded": {}, "failed": {}}
    for i, item_submission in enumerate(self.item_submissions_iter()):
        item_id = item_submission.item_identifier
        try:
            item_submission.upload_dspace_metadata(self.s3_bucket, self.batch_path)
            response = item_submission.send_submission_message(
                self.workflow_name,
                self.output_queue,
                self.submission_system,
                self.collection_handle,
            )
            status_code = response["ResponseMetadata"]["HTTPStatusCode"]
            if status_code != HTTPStatus.OK:
                items["failed"][item_id] = RuntimeError("Non OK HTTPStatus")
                continue
            items["succeeded"][item_id] = response["MessageId"]
        except Exception as exc:
            logger.exception(f"Error processing submission: {item_id}")
            items["failed"][item_id] = exc

    results = {
        "success": not items["failed"],
        "items_count": len(items["succeeded"]) + len(items["failed"]),
        "items": items,
    }
    logger.info(results)
    return results
```

But I acknowledge this is all very subjective.  What you have originally certainly works!  I'm coming at this from the POV of debugging or interactively running this: what is my ideal data structure returned?  

Personally, I would love a list of successful items (knowing the value associated with each item identifier is a SQS message ID) and a list of failed items (knowing the value associated with the item identifier is a true `Exception` object that contains the message and full stacktrace).

ghukill (commented):
Changes are looking good.  Left some stylistic / data structure comments about the new `Workflow.run()` method, but please do elect to ignore or pushback.  As noted in the 2nd comment, it's subjective; I believe what is currently in place would work.

Leaving a "Comment" review, in the event you were considering adding the Workflow init changes to this PR.  Otherwise, bordering on "Approve"!

ghukill (commented):
Whoops, meant to leave a "Comment" review.

ehanson8:
Had a fruitful huddle with @ghukill and @jonavellecuerdo and pushed a new commit based on that discussion, thanks again!

jonavellecuerdo (approved):
Awesome work! ✨ Thank you for your openness and communication across several iterations of the code in this PR. I think it ended in a good place, and we, as a team, overall have a better understanding of how this CLI app should work. 🤓 

Side note: Just one minor comment about the type for the `email_recipients` param!

dsc/workflows/__init__.py
```diff
@@ -3,8 +3,8 @@
...
 
-from dsc.workflows.base import BaseWorkflow
+from dsc.workflows.base import Workflow
+from dsc.workflows.base.simple_csv import SimpleCSV
+from dsc.workflows.demo import DemoWorkflow
 
-__all__ = [
-    "BaseWorkflow",
-]
+__all__ = ["DemoWorkflow", "SimpleCSV", "Workflow"]
```
ghukill, line 10: I didn't comment on this earlier, but I continue to really like this.  Not only does it import and therefore register all workflows as a `__subclass__` of `Workflow`, but it allows for this:

```python
from dsc.workflows import DemoWorkflow
```

ghukill (approved):
Looks good to me.  

Really appreciate the facilititating of discussions @ehanson8, and the plasticity to incorporate discussion points with pre-existing structures.  It _is_ funny how old is new again (simple `if` blocks to set instance values, as it was originally proposed).

Most of all, I feel like I can drift around the code and mostly reason about what's happening and why.  Nice work!

dsc/cli.py
```diff
@@ -66,11 +66,11 @@ def main(
...
     ctx.ensure_object(dict)
     ctx.obj["start_time"] = perf_counter()
-    workflow = Workflow.load(
-        workflow_name=workflow_name,
+    workflow_class = Workflow.get_workflow(workflow_name)
+    workflow = workflow_class(
         collection_handle=collection_handle,
         batch_id=batch_id,
-        email_recipients=email_recipients,
+        email_recipients=tuple(email_recipients.split(",")),
```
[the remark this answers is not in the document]
ehanson8: I did try a list initially but that opened a whole mutable-types can-of-worms re: [RUF012](https://docs.astral.sh/ruff/rules/mutable-class-default/) so I'm keeping as a tuple for now but this may change in the `finalize` PR since we'll actually be using them

PR #91, opened 2025-01-23
In 1096 ocw workflow

jonavellecuerdo:
This PR creates the `OpenCourseWare` DSC workflow.

**A. Review the added unit tests.**

NO

NO

- https://mitlibraries.atlassian.net/browse/IN-1096

### Developer

### Code Reviewer(s)


dsc/workflows/opencourseware.py
```diff
@@ -0,0 +1,158 @@
...
+        ).strip()
+
+    @staticmethod
+    def _construct_instructor_name(instructor: dict[str, str]) -> str:
+        """Given a dictionary of name fields, derive instructor name."""
+        if not (last_name := instructor.get("last_name")) or not (
+            first_name := instructor.get("first_name")
+        ):
+            return ""
+        return f"{last_name}, {first_name} {instructor.get("middle_initial", "")}".strip()
```
jonavellecuerdo, line 138: While it is plausible that all the metadata in `data.json` will always be formatted as needed (i.e., all `instructor` name fields provided), it would be a good idea to check in with stakeholders on the "minimum required `instructor` name fields` to construct an instructor name.

In this sample mapping file we received, [ocw_json_to_dspace_mapping.xlsx](https://github.com/user-attachments/files/18525047/ocw_json_to_dspace_mapping.xlsx), it indicates the `instructor` names must be formatted as: 

```text
<last_name>, <first_name> <middle_initial>
```

The code above will return an empty string if either the `last_name` or `first_name` is missing; it allows for missing `middle_initial` values.

jonavellecuerdo:
Met with @ghukill last Friday and wanted to share some thoughts from our discussion: 

1. **Clarify language around the `reconcile` method**: The base `Workflow.reconcile_bitstreams_and_metadata` method uses the terms "item_identifiers_without_bitstreams" and "bitstreams_without_item_identifiers". The use of these terms felt a bit awkward with the `OpenCourseWare` workflow for the following reason:

   > . . . the 'reconcile' method only determines whether there are any bitstreams without metadata (any zip files without a 'data.json'). Metadata without bitstreams are basically impossible because the metadata ('data.json') is inside the bitstream (zip file)
   
   My interpretation of `reconcile` is that it is really about determining whether there is metadata provided 
3.

ehanson8:
Agree that 1 & 2 should be handled as separate tickets!

tests/test_workflow_opencourseware.py
```diff
@@ -0,0 +1,240 @@
...
+        opencourseware_source_metadata,
+        FileNotFoundError,
+    ]
+
+    assert opencourseware_workflow_instance._identify_bitstreams_with_metadata(
+        item_identifiers=["123", "124"]
+    ) == ["123"]
+
+
+@patch("dsc.workflows.opencourseware.OpenCourseWare._read_metadata_json_file")
```
ehanson8, line 88: Both of these tests could be a lot simpler with 2 zip file fixtures (one with metadata, one without) that you load into `mocked_s3` with the `s3_client` fixture as you did in the `test_workflow_ocw_build_bitstream_dict_success`

tests/test_workflow_opencourseware.py
```diff
@@ -0,0 +1,240 @@
...
+    ) == ""
+
+    opencourseware_workflow_instance._construct_instructor_name(
+        {"first_name": "Kerry", "middle_initial": ""}
+    ) == ""
+
+    opencourseware_workflow_instance._construct_instructor_name(
+        {"last_name": "Oki", "middle_initial": ""}
+    ) == ""
+
```
ehanson8, line 221: Great set of comprehensive tests for the instructor name methods!

ehanson8 (changes requested):
Looking great but a few requested changes!

dsc/workflows/opencourseware.py
```diff
@@ -0,0 +1,203 @@
...
+        s3_client = S3Client()
+        bitstreams = list(
+            s3_client.files_iter(
+                bucket=self.s3_bucket, prefix=self.batch_path, file_type=".zip"
+            )
+        )
+        bitstream_dict: dict[str, list[str]] = defaultdict(list)
+        for bitstream in bitstreams:
+            file_name = bitstream.split("/")[-1]
+            item_identifier = file_name.replace(".zip", "")
```
ghukill, line 70: Relatively minor and unlikely to crop up, but I'd be cautious about using `.replace()` in situations like this.  If the filename was something like `mr.zips-amazing-class.zip` you'd get `mrs-amazing-class` as the identifier because both instances of `.zip` were removed:

```python
x = "mr.zips-amazing-class.zip"
x.replace(".zip", "")
Out[2]: 'mrs-amazing-class'
```

I'd suggest something like `.split(".")[0]` but that is also problematic if there is a period `.` anywhere in the filename.

SO, taken altogether, feels like `.removeprefix(".zip")` could be the best bet here.

dsc/workflows/opencourseware.py
```diff
@@ -0,0 +1,203 @@
...
+
+    def _identify_bitstreams_with_metadata(
+        self, item_identifiers: list[str]
+    ) -> list[str]:
+        bitstreams_with_metadata = []
+        for item_identifier in item_identifiers:
+            try:
+                zip_file = (
+                    f"s3://{self.s3_bucket}/{self.batch_path}/{item_identifier}.zip"
+                )
```
ghukill, line 82: What if `self._build_bitstream_dict()` returned a tuple like `(<indentifier>, <bitstream_filename>, <bitstream_fullpath>)`?  If you had the full path of the bitstream returned there, when discovered, you could theoretically pass that list of tuples to _this_ method and it wouldn't have to construct the path here.

I certainly see how this works, but feels potentially brittle and/or confusing why we use `s3_client.files_iter(...)` with parameters in one place, but we manually construct S3 URIs here.

You could do something like:
```python
for item_identifier, _, bitstream_uri in bitstreams:
   # do stuff
```

where `bitstreams` is this slightly modified list of bitstreams in S3 that were discovered via `self._build_bitstream_dict()`.

ghukill (commented):
Overall, looking good to me.  Thanks for the discussion the other day, which was quite helpful.

I left a couple of comments/suggestions for fairly minor, syntactical things.  None are required or blocking.

I did have another comment that this PR surfaced for me.  Start by saying perhaps this is could be another ticket for exploring.

Should the CLI command `reconcile` potentially return a non-zero exit code?  Thinking forward to automation, throwing an exit code like `1` or `2` would indicate that reconciliation had failed in some way.  This could be helpful for humans, but more helpful for automation like StepFunctions.  

I think this came to mind in this PR given the conversations about reconciling and what it's conceptually communicating.

dsc/workflows/opencourseware.py
```diff
@@ -0,0 +1,203 @@
...
+
+    workflow_name: str = "opencourseware"
+    metadata_mapping_path: str = "dsc/workflows/metadata_mapping/opencourseware.json"
+
+    def reconcile_bitstreams_and_metadata(self) -> tuple[set[str], set[str]]:
+        """Reconcile bitstreams against metadata.
+
+        Generate a list of bitstreams without item identifiers and item identifiers
+        without bitstreams. Any discrepancies will be addressed by the engineer and
+        stakeholders as necessary.
```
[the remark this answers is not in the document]
ghukill: Had the same thought.  I think it actually gets at kind of a deeper pattern about docstrings and inheritance.  I think we're _mostly_ pretty good about docstrings hitting the balance of helpful // won't go stale too quickly.   But in this case, agreed: dupliating docstrings from parent classes feels like something we'd probably want to avoid.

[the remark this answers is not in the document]
jonavellecuerdo: I updated the docstring such that references to 'metadata' -> 'item metadata'. I also updated the second line to instead read:

> Generate a list of bitstreams without item metadata.

dsc/workflows/opencourseware.py
```diff
@@ -0,0 +1,203 @@
...
+                prefix=self.batch_path,
+                item_identifier=item_identifier,
+                file_type=".zip",
+                exclude_prefixes=["archived"],
+            )
+        )
+
+    def process_deposit_results(self) -> list[str]:
+        """TODO: Stub method."""
+        return [""]
```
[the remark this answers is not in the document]
jonavellecuerdo: To be addressed via ticket [IN-1156](https://mitlibraries.atlassian.net/browse/IN-1156).

tests/test_workflow_opencourseware.py
```diff
@@ -0,0 +1,240 @@
...
+        "middle_initial": "E.",
+        "salutation": "Prof.",
+        "title": "Prof. Earl E. Bird",
+    },
+]
+
+
+@patch("dsc.workflows.opencourseware.OpenCourseWare._extract_metadata_from_zip_file")
+@patch("dsc.utilities.aws.s3.S3Client.files_iter")
+def test_workflow_ocw_reconcile_bitstreams_and_metadata_success(
```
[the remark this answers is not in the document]
jonavellecuerdo: This test actually makes use of a list of side effects to demonstrate what happens if a batch comprises of one zipped file _with_ item metadata and one zipped file _without_ item metadata.

In addition to verifying that the second list (the set of "bitstreams without item metadata") returned by `OpenCourseWare.reconcile_bitstreams_and_metadata` indicates the second item (for which a `FileNotFoundError` is raised), `{124}`, I included an assertion that the expected error message appears in `caplog`. 

TLDR: Included one more assertion instead of writing a new unit test! 😅

dsc/workflows/opencourseware.py
jonavellecuerdo: Ah, thank you for this suggestion! I will use these methods (`removesuffix` in this case) moving forward.

tests/test_workflow_opencourseware.py
jonavellecuerdo: Changes made. See [updates](https://github.com/MITLibraries/dspace-submission-composer/blob/IN-1096-ocw-workflow/tests/test_workflow_opencourseware.py#L77-L106).

dsc/workflows/opencourseware.py
jonavellecuerdo: Please review the [latest version of `OpenCourseWare.reconcile_bitstreams_and_metadata`](https://github.com/MITLibraries/dspace-submission-composer/blob/IN-1096-ocw-workflow/dsc/workflows/opencourseware.py#L27-L62) and let me know what you think! Changes introduced in PR #102 allowed more flexibility when defining the 'reconcile' method for OCW.

dsc/workflows/opencourseware.py
```diff
@@ -0,0 +1,183 @@
...
+        exist, the zip file must also exist.
+        """
+        item_identifiers = []
+        bitstreams_without_metadata = []
+        s3_client = S3Client()
+        for file in s3_client.files_iter(
+            bucket=self.s3_bucket, prefix=self.batch_path, file_type=".zip"
+        ):
+            zip_file = f"s3://{self.s3_bucket}/{file}"
+            item_identifier = file.split("/")[-1].removesuffix(".zip")
```
jonavellecuerdo, line 46: Note: This function contains similar steps to `OpenCourseWare.item_metadata_iter`; only, instead of yielding a dictionary of metadata, it checks whether metadata can be extracted from the zip file--tracking item identifiers in two lists:
1. `item_identifiers`: list of item identifiers for all zip files
2. `bitstreams_without_metadata`: list of item identifiers for all zip files without a 'data.json' file.

dsc/workflows/opencourseware.py
```diff
@@ -0,0 +1,183 @@
...
+    ) -> dict[str, str]:
+        """Read source metadata JSON file."""
+        with zip_file.open(metadata_json) as file:
+            source_metadata = json.load(file)
+            source_metadata["instructors"] = self._get_instructors_delimited_string(
+                source_metadata["instructors"]
+            )
+            return source_metadata
+
+    def _get_instructors_delimited_string(self, instructors: list[dict[str, str]]) -> str:
```
ghukill, line 117: Noting @jonavellecuerdo's [comment #2 here](https://github.com/MITLibraries/dspace-submission-composer/pull/91#issuecomment-2615913965), around exploring if multi-value lists/arrays can be passed for metadata parsing, to avoid creating a delimited string just to parse it again.

I would be in favor of either an issue or a new ticket and _not_ incorporate that in this PR.  But noting here, as I think this PR exposes a situation when we're creating delimited strings from structured data only to re-structure it moments later.

ghukill (changes requested):
Think it's looking great!

My only request is a standalone method for the item identifier parsing, to ensure it's always done the same way for any part of the workflow that attempts to establish it for an item.

dsc/workflows/opencourseware.py
ehanson8: Agree that a ticket would be the best approach

dsc/workflows/opencourseware.py
```diff
@@ -24,67 +24,47 @@ class OpenCourseWare(Workflow):
...
-
-        This method will look for zip files within the designated 'batch' folder
-        of the S3 bucket (i.e., self.batch_path). In the case of OpenCourseWare
-        deposits, this method expects:
-            * a single (1) zipped file per item identifier
-            * filename of zipped file must correspond to item identifier
-              (i.e., '<item_identifier>'.zip)
+        bitstreams without metadata (any zip files without a 'data.json').
+        Metadata without bitstreams is not calculated as for a 'data.json' file to
+        exist, the zip file must also exist.
```
ehanson8, line 37: Good docstring!

dsc/workflows/opencourseware.py
ehanson8: @ghukill Did you mean these 2 lines? If so, I agree

```
zip_file = f"s3://{self.s3_bucket}/{file}"
item_identifier = file.split("/")[-1].removesuffix(".zip")
```

ehanson8 (commented):
Looks good to me, I concur with Graham's method and ticket suggestions and then full approval!

dsc/workflows/opencourseware.py
jonavellecuerdo: Ticket has been created [IN-1159](https://mitlibraries.atlassian.net/jira/software/c/projects/IN/boards/238?selectedIssue=IN-1159). :)

dsc/workflows/opencourseware.py
jonavellecuerdo: Let me know if [these changes](https://github.com/MITLibraries/dspace-submission-composer/pull/91/commits/3ca48aa313858b55eb1e0474db03cc268abb91c7) are in line with what you were thinking, @ghukill and @ehanson8 !

jonavellecuerdo:
Had to reapply the changes to rename the test module for `SimpleCSV` workflow via commit https://github.com/MITLibraries/dspace-submission-composer/pull/91/commits/7abbc7cab9ae797748029e09d5c92e9f9793a7e4. I dropped the first version of the commit from this branch to avoid rebasing issues with `main` (since new tests were added to `SimpleCSV` test module).

dsc/workflows/opencourseware.py
```diff
@@ -76,11 +75,10 @@ def item_metadata_iter(self) -> Iterator[dict[str, Any]]:
         for file in s3_client.files_iter(
             bucket=self.s3_bucket, prefix=self.batch_path, file_type=".zip"
         ):
-            zip_file = f"s3://{self.s3_bucket}/{file}"
-            item_identifier = file.split("/")[-1].removesuffix(".zip")
+            item_identifier = self.parse_item_identifier(file)
             yield {
-                "item_identifier": item_identifier,
-                **self._extract_metadata_from_zip_file(zip_file, item_identifier),
+                "item_identifier": self.parse_item_identifier(file),
```
ghukill, line 80: Sorry to nit-pick, but I think you have `item_identifier` in hand a couple of lines up, can this just be used here?

ghukill (approved):
Looks good from my POV, even with a nit-pick comment about using a variable rather then re-calling a method.

Slowing down on this PR and working on the reconcile methods, and backlogging some tickets, really feels like the right approach.  Hat tip to all involved here for moving the project along in a comprehensive way.

dsc/workflows/opencourseware.py
jonavellecuerdo: Good catch! Fixed.

PR #121, opened 2025-02-06
Build framework for creating reports

jonavellecuerdo:
The app requires a framework for creating reports that summarize results from DSC workflow runs. This framework should support creating different reports for the three main DSC CLI commands: `reconcile`, `submit`, and `finalize`. To demonstrate the construction of a report, this PR adds a `FinalizeReport`, which represents the email that is sent when the `finalize` CLI command is executed. The `FinalizeReport` produces an email with the following: 

* [Subject heading](https://github.com/MITLibraries/dspace-submission-composer/blob/IN-1157-build-report-system/dsc/report.py#L101-L105): "DSpace Submission Results - <workflow_name>, batch='<batch_id>'"
* [File attachments](https://github.com/MITLibraries/dspace-submission-composer/blob/IN-1157-build-report-system/dsc/report.py#L115-L138): a CSV file describing successfully deposited items (columns for `item_identifier` and `doi`) and a text file for captured error messages (from processing DSS results)
* Message body: See text in 'finalize' [HTML](https://github.com/MITLibraries/dspace-submission-composer/blob/IN-1157-build-report-system/dsc/templates/html/finalize.html) and [plain-text](https://github.com/MITLibraries/dspace-submission-composer/blob/IN-1157-build-report-system/dsc/templates/plain_text/finalize.txt) templates.

This PR touches several files, so here are some notes for reviewers:
* One small change in the `Makefile` is to fix the `lint-apply` syntax (to run multiple `Make` commands, it seems they must be on the same line.
* The new `dsc/templates` contain 4 very simply formatted HTML and plain-text templates. There is a 'base' template that should be extended by templates for 'child' reports (i.e., report templates for `reconcile`, `submit`, and `finalize`). 
* Recommend reviewing from top-level to low-level (i.e., start with the `cli` then dig 🤓 ). See order of changes described in commit message.

* Review [unit tests in `test_report.py`](https://github.com/MITLibraries/dspace-submission-composer/pull/121/files#diff-59aa45b5e4107d2c022be08538060a2f3701cfe1657eadb979ef9a077cdf8193)
* Review [updated unit test in `test_cli.py`](https://github.com/MITLibraries/dspace-submission-composer/pull/121/files#diff-4e8715c7a425ee52e74b7df4d34efd32e8c92f3e60bd51bc2e1ad5943b82032e) demonstrating use of new boolean flag option
* Review [updated unit tests in `test_ses.py`](https://github.com/MITLibraries/dspace-submission-composer/pull/121/files#diff-6a5cd2f3d8b2103acf9d2b36632c31a5d4ef5885152287969c32bd356e98f09d) demonstrating changes to include email message body + support new format of `attachments` arg

**Note:** I contemplated whether to add unit tests to verify the rendering of the report templates, but I would prefer to review the resulting email once we have the AWS infrastructure in place to send and review test emails. Once that is in place, I plan on including screenshots to this PR/or the ticket for reference. However, this shouldn't block the PR from being merged!

YES - installs `jinja2`; installs `freezegun` as 'dev-package'.

YES - As of this writing, the existing report templates are quite simple,
with the goal of getting to an minimum viable product. It would be good
to include in coming documentation what email recipients can expect
in the emailed report.

https://mitlibraries.atlassian.net/browse/IN-1157

### Developer

### Code Reviewer(s)


dsc/cli.py
```diff
@@ -27,20 +28,33 @@
...
+    default=False,
+)
 @click.option(
     "-v", "--verbose", is_flag=True, help="Pass to log at debug level instead of info"
 )
 def main(
     ctx: click.Context,
     workflow_name: str,
     batch_id: str,
+    create_and_send_report: bool,  # noqa: FBT001
```
ehanson8, line 49: Is there a reason to put this on `main` rather than the CLI command itself? Then there would be no need to add it to the context object

dsc/cli.py
```diff
@@ -27,20 +28,33 @@
...
     required=True,
 )
+@click.option(
+    "-r",
+    "--create-and-send-report",
+    is_flag=True,
+    help=(
+        "Pass to create and send summary report to recipients. "
+        "Reports generated for 'reconcile' and 'submit' steps must only be sent to "
+        "DSC admins."
```
ehanson8, line 38: Even though I'd like to see this param shift to the specific CLI commands, I would also be less specific about who it is sent to, in `wiley` for example, the `submit` report would be sent to stakeholders as well. The first line seems sufficient

dsc/cli.py
```diff
@@ -117,4 +131,8 @@ def finalize(ctx: click.Context, email_recipients: str) -> None:
     """Process the result messages from the DSS output queue according the workflow."""
     workflow = ctx.obj["workflow"]
     workflow.process_results()
-    workflow.report_results(email_recipients.split(","))
+
+    if ctx.obj["create_and_send_report"]:
```
ehanson8, line 135: I think this boolean check is better suited to `submit`, I have a hard time imagining a workflow where the app wouldn't report out the results of the DSS run. 

Also, if the `create_and_send_report` param is moved to `submit`, it could look this:

```
if create_and_send_report:
    workflow.send_report(
            report_class=FinalizeReport, email_recipients=email_recipients.split(",")
        )
```

And it could probably also be simplified to just `report`

dsc/report.py
```diff
@@ -0,0 +1,190 @@
...
+        if self.events.processed_items and not self.events.errors:
+            return "success"
+        if self.events.processed_items and self.events.errors:
+            return "incomplete"
+        return "error"
+
+    def create_attachments(self) -> list[tuple]:
+        """Create file attachments for 'finalize' email.
+
+        This method will create a CSV file of successfully depposited
```
ehanson8, line 118: Typo on "successfully deposited"

dsc/report.py
ehanson8, line 1: I really like your approach here, this is a very elegantly-design module!

dsc/report.py
```diff
@@ -0,0 +1,190 @@
...
+
+        This method creates a string buffer with the error messages
+        encountered when 'finalize' methods were executed.
+        """
+        errors = self.events.errors
+
+        text_buffer = StringIO()
+        for error in errors:
+            text_buffer.write(error + "\n")
+        text_buffer.seek(0)
```
ehanson8, line 167: A single report could allow for some simplification here as well

dsc/workflows/base/__init__.py
```diff
@@ -19,10 +19,29 @@
...
+    This dataclass is designed to hold useful data used in reporting.
+    It is comprised of three lists, which contain details
+    about reconciled, submitted, and processed items -- aligning with
+    the DSC CLI commands (reconcile, submit, and finalize). Error
+    messages are also tracked in a list.
+    """
+
+    reconciled_items: list[str] = field(default_factory=list)
+    submitted_items: list[str] = field(default_factory=list)
+    processed_items: list[dict] = field(default_factory=list)
```
ehanson8, line 41: `finalize` vs `processed` injects some ambiguity and I'm not sure which is preferable, but DataEng should probably decide on that for consistency to prevent confusion

dsc/workflows/base/__init__.py
```diff
@@ -19,10 +19,29 @@
...
+@dataclass
+class WorkflowEvents:
+    """Record of events during the execution of Workflow methods.
+
+    This dataclass is designed to hold useful data used in reporting.
+    It is comprised of three lists, which contain details
+    about reconciled, submitted, and processed items -- aligning with
+    the DSC CLI commands (reconcile, submit, and finalize). Error
+    messages are also tracked in a list.
+    """
```
ehanson8, line 37: I like this!

dsc/report.py
```diff
@@ -0,0 +1,190 @@
...
+    def to_rich_text(self) -> str:
+        template = self.env.get_template(f"{self.template_name}.html")
+        return template.render(
+            workflow_name=self.workflow_name,
+            batch_id=self.batch_id,
+            report_date=self.report_date,
+            status=self.status,
+            processed_items=self.events.processed_items,
+            errors=self.events.errors,
+        )
```
ehanson8, line 190: I like the flexibility of txt, html, and/or attachments but am I interpreting this correctly that the same successes and errors would be duplicated in all 3 formats?

dsc/workflows/base/__init__.py
```diff
@@ -42,7 +61,7 @@ def __init__(
                 and metadata files.
         """
         self.batch_id = batch_id
-        self.report_data: list[str] = []
+        self.workflow_events = WorkflowEvents()
```
ehanson8, line 64: Great rename!

tests/test_report.py
```diff
@@ -0,0 +1,85 @@
+from unittest.mock import patch
+
+from freezegun import freeze_time
+
+from dsc.report import FinalizeReport
+from dsc.workflows.base import WorkflowEvents
+
+
+@freeze_time("2025-01-01 09:00:00")
+def test_finalize_report_init_success(workflow_events_finalize):
```
ehanson8, line 10: Would we ever instantiate a report this way or is it always expected to be `.from_workflow`?

tests/test_report.py
```diff
@@ -0,0 +1,85 @@
...
+        workflow_name="test", batch_id="aaa", events=workflow_events_finalize
+    )
+    assert finalize_report.subject == "DSpace Submission Results - TEST, batch='aaa'"
+
+
+def test_finalize_report_status_returns_success(workflow_events_finalize):
+    finalize_report = FinalizeReport(
+        workflow_name="test", batch_id="aaa", events=workflow_events_finalize
+    )
+    assert finalize_report.status == "success"
```
ehanson8, line 44: Could these tests be combined?

ehanson8 (commented):
This is really great, a number of questions and comments but this was easy to read and understand how it integrates into the app. Fantastic work

dsc/workflows/base/__init__.py
```diff
@@ -300,33 +319,42 @@ def process_sqs_queue(self) -> list[tuple[str, dict]]:
...
-                self.report_data.append(f"{item_identifier}: {message_body}")
+                items.append(
+                    {"item_identifier": item_identifier, "message_body": message_body}
+                )
+                self.workflow_events.processed_items.append(
+                    {
+                        "item_identifier": item_identifier,
+                        "doi": message_body["ItemHandle"],
+                    }
+                )
```
ghukill, line 330: These two objects look _very_ similar.  This also goes back to an interesting discussion I recall having with @ehanson8 about passing data vs reporting.

For the `WorkflowEvents`, do we _want_ to accumlate objects like this?  Or might we want more human-friendly strings?

**Argument for objects**: this gives the report the ability to change what it retrieves from the object later, and how it displays that.

**Argument for strings**: the workflow is deciding how to "spin" or describe the event, and it's a static string that it provides.

I don't think there is a "correct" answer, just an opinionated one.  But I do think answering that question has bearing on how these two objects look very similar.

dsc/cli.py
```diff
@@ -117,4 +131,8 @@ def finalize(ctx: click.Context, email_recipients: str) -> None:
...
+        workflow.send_report(
+            report_class=FinalizeReport, email_recipients=email_recipients.split(",")
+        )
```
ghukill, line 138: As mentioned in my first PR review summary, I reflected on this more after typing out the big moving pieces.

I wonder if `Workflow.send_report()` should be unaware / not require a `Report` instance.  If instead you provided arguments things like:
- `message_body_plaintext`
- `message_body_html`
- `attachments`
- etc.

Seems like it would make `send_report()` easier to test, as it's just static values that get used.  It keeps the separation of concerns stronger, as Workflows can still _fully_ function without the notion of a `Report` object.  

Not a hard requirement, but posing as a question to see how it feels.

ghukill (changes requested):
As a first pass, some questions and requests for discussion.

Overall, think it looks great!  I had the opportunity to discuss and see sketches of this via a huddle, so felt _very_ easy to reconigze how it would work.  But I think that would be true even without pre-exposure.

I break it down into these big blocks:
- Workflows _always_ capture noteworthy events that happened while they worked, for reconcile, submit, or finalize
- Reports can _optionally_ be created after a workflow completes, but using data from the complete workflow object (name, events, etc.)
- Reports can serialize themselves into plain text or HTML
- Workflows have the ability to "report out", so it's their method `send_report` that is used

As I type all this... I wonder if Workflow's `send_report()` should actually be unaware of a `Report` object?  Adding as a final comment in my review, only surfaced from this typing here. 

But, overall, looking real good.

dsc/cli.py
ghukill: Heartily agree.  I _almost_ commented elsewhere about docstring that was _almost_ too specific, making it somewhat brittle over time.  I do like detailed docstrings, but erring on the side of describing unusual mechanics over details that are subject to change.

jonavellecuerdo: Updated.

dsc/workflows/base/__init__.py
jonavellecuerdo: Shoutout to @ghukill for this idea!

dsc/report.py
ghukill: I'll chime in that at least duplicating between rich text (really HTML) and plain text is normal for emails.  When someone reads the email they will only see the rich text _or_ the plain text version.

But I think your point stands if the summary is showing successes/errors _and_ the standalone attachments are as well.

I'd want to double check, but I wonder if the report summaries are only counts?  and don't include actual item identifiers or specific errors?

tests/test_report.py
jonavellecuerdo: Hmm, I could use some clarification:
* `test_finalize_report_subject_success` verifies the **subject heading** set for the email for a 'finalize' report 
* `test_finalize_report_status_returns_success` verifies the **status** included in the email message body for a 'finalize' report (see screenshot below for sample message body).

image

dsc/report.py
```diff
@@ -0,0 +1,190 @@
...
+            )
+
+        if self.events.errors:
+            attachments.append(
+                (
+                    "errors.txt",
+                    self._write_errors_text_file(),
+                )
+            )
+        return attachments
```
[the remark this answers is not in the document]
jonavellecuerdo: 🤔

tests/test_report.py
ehanson8: OK, sorry, I misinterpreted what's happening here, totally fine as is, thank you for clarifying!

dsc/report.py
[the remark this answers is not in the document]
ehanson8: I like that approach, it would mirror what we're doing in the `workflows` directory

dsc/report.py
```diff
@@ -0,0 +1,190 @@
...
+    ):
+        self.workflow_name = workflow_name.upper()
+        self.batch_id = batch_id
+        self.report_date = datetime.datetime.now(tz=datetime.UTC).strftime(
+            "%Y-%m-%d %H:%M:%S"
+        )
+        self.events = events
+
+        # configure environment for loading jinja templates
+        self.env = Environment(
```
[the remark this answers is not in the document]
ehanson8: Agreed!

dsc/report.py
```diff
@@ -0,0 +1,190 @@
...
+        )
+
+    def to_rich_text(self) -> str:
+        template = self.env.get_template(f"{self.template_name}.html")
+        return template.render(
+            workflow_name=self.workflow_name,
+            batch_id=self.batch_id,
+            report_date=self.report_date,
+            status=self.status,
+        )
```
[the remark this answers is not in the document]
ehanson8: I think this is a fine default that would work for most reports so not sure about the `@abstractmethod`, but certainly let it be overridden by subclasses as necessary

dsc/report.py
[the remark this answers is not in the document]
ehanson8: Taking another parallel from the workflow classes, could `self.env.get_template("hard_coded_template_name}.txt")` be a property that is set in the subclasses but called here?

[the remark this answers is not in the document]
ghukill: I like that pattern @ehanson8.  And maybe could extend one hop further.  Maybe the properties are the actual template filename, which allows leaving `self.env.get_template(...)` the same for all classes, and therefore set as a property on the base `Report` class?

Like:

```python
class Report:
    @property
    @abstractmethod
    def jinja_template_plain_text_filename(self): ...

    @property
    @abstractmethod
    def jinja_template_html_filename(self): ...

    @property
    def jinja_template_plain_text(self):
        return self.env.get_template(self.jinja_template_plain_text_filename)

    @property
    def jinja_template_html(self):
        return self.env.get_template(self.jinja_template_html_filename)


class FinalReport(Report):
    @property
    def jinja_template_plain_text_filename(self):
        return "final.txt"

    @property
    def jinja_template_html_filename(self):
        return "final.html"
```

jonavellecuerdo: Good questions here! Some points of clarification:

* The HTML and plain-text templates outline the content of the **email message body** (see screenshot)
   image
   * What I interpreted in [this article, _Sending Emails With Python_](https://realpython.com/python-send-email/#sending-fancy-emails), is that when providing an HTML template, it's good to include a plain-text file in case the email client can't display HTML content for whatever reason. 🤔 
   * The screenshot above shows what the rendered HTML content from `dsc/finalize.html` would look like.
   * The rendered plain-text content `dsc/templates/plain_text/finalize.txt` would be similar, without any text formatting (i.e., the first line would not be bolded.
   * @ghukill , the message body only displays the counts
* The attachments for a 'finalize' report are:
   * a CSV file with columns for `file_identifier` and `doi`
   * a text file with all error messages (i.e., [these error messages](https://github.com/MITLibraries/dspace-submission-composer/blob/IN-1157-build-report-system/dsc/workflows/base/__init__.py#L333))

@ehanson8 , I don't think there is duplication, but I will be curious to see what the emails look like once we can start sending test emails! I'm fairly new to writing email-sending functions, but the first bullet above is how I understand the current setup to work. 🤔

ehanson8: Got it, that all sounds fine, we can discuss more tomorrow!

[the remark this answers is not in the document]
ehanson8: I like that extension a lot!

dsc/report.py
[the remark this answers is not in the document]
jonavellecuerdo: [IN PROGRESS] 

That is a fair question! I am a bit hesitant for something completely different (in the spirit of moving forward), _but_ I am definitely on the camp of: yes, let's continue to think on how we can make this better but not get too bogged down trying to make it super awesome because I definitely think ideas will come as this app begins to see some use, and we start getting some feedback from stakeholders / maybe after reviewing some test emails). 🤔 

That is to say, I think it boiled down to these files because prior to these changes, it was originally creating a text file that could look like this...

dsc/report.py
```diff
@@ -0,0 +1,190 @@
...
+    @property
+    @abstractmethod
+    def subject(self) -> str:
+        """Subject heading used in report email."""
+
+    @classmethod
+    def from_workflow(cls, workflow: workflows.Workflow) -> Report:
+        """Create instance of Report using dsc.workflows.Workflow."""
+        return cls(
+            workflow_name=workflow.workflow_name.upper(),
```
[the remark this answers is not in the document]
jonavellecuerdo: ...Because SCREAMING-SNAKE 👀 Heh

I can undo this since I wasn't sure! I can definitely use some input/design advice on how to structure the email message body. I wanted to make it clear which workflow was run. 😅

dsc/report.py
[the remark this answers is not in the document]
jonavellecuerdo: Great suggestion! Updated.

dsc/cli.py
jonavellecuerdo: Hmm, to clarify, are you suggesting the code shown below? 🤔:

```python
def finalize(ctx: click.Context, email_recipients: str) -> None:
    """Process the result messages from the DSS output queue according the workflow."""
    workflow = ctx.obj["workflow"]
    workflow.process_results()
    report = FinalizeReport.from_workflow(workflow)
    
    if ctx.obj["create_and_send_report"]:
        workflow.send_report(
           subject=report.subject,
           message_body_plaintext=report.to_plaintext(),
           message_body_html=report.to_html(), 
           email_recipients=email_recipients.split(",")
        )
```

Looking at the above, I think I still prefer the current structure, as it:
* nudges a developer in the direction of the `Report` class and `jinja` templates to structure the emails; these two encourage consistency in reporting + better track different reports / emails by way of templates.
* reduces the number of arguments on `Workflow.send_report`; the proposed argument `message_body_html` seems like would be a bit difficult (or maybe awkward) to provide (i.e., providing a hardcoded HTML string)

Let me know what you think!

dsc/workflows/base/__init__.py
jonavellecuerdo: For the reasons shared above, I think I prefer **objects**. It seems like it would be easier to either create a custom report class or update a template vs. changing a string in `WorkflowEvents` (and in `Workflow` method calls where `WorkflowEvents` is populated) whenever a rewording is needed. Similarly, adding a data point to `WorkflowEvents` if something is missing and then updating the report template feels more flexible...

What do you think, @ehanson8 and @ghukill ? 🤔

dsc/report.py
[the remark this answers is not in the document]
jonavellecuerdo: Resolving this comment in favor implementing the proposal in https://github.com/MITLibraries/dspace-submission-composer/pull/121#discussion_r1945505830.

dsc/report.py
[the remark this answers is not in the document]
jonavellecuerdo: Also wanted to remind that the error messages are captured as-is in a text file `errors.txt` that is included as an attachment in the email.

dsc/workflows/base/__init__.py
jonavellecuerdo: Ah, the age-old question! 😅 

My interpretation:
> The `finalize` CLI command performs two functions: **processing** the DSS result messages sent to the output queue for a given workflow and sending the report (an email with attachments) to designated recipients. For the report, we are interested in knowing what DSS result messages sent to the output queue were _successfully_ **processed** (i.e., a valid DSS result message means it was successfully submitted to DSpace).

With that understanding, I think `processed_items` is a reasonable name for this collected data, but I am flexible!

tests/test_report.py
jonavellecuerdo: Good question! I think, yes, it would always be `.from_workflow`, but I think the `__init__` method can be handy for testing?

dsc/report.py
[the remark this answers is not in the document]
jonavellecuerdo: Thank you for the suggestion! Updated.

dsc/workflows/base/__init__.py
ghukill: @jonavellecuerdo - objects make sense to me!  I guess my followup regarding their _near_ similarity: why not use the same object for both then?  It looks like the reporting  object just  is a _subset_ of the SQS message body, but if you pass the whole thing, then reporting has more options on what to selectively report on.

dsc/cli.py
ghukill: @jonavellecuerdo - FWIW, yeah, that's about what I was thinking!  Possibly capturing each string value in a variable first, then passing to `workflow.send_report()` (for debugging and clarity sake), but that's a minor point.

Ultimately, I think I agree with your two bullets.  I'd pushback a little that providing HTML is difficult,

```python
workflow.send_report(
   subject="Hello World!",
   message_body_plaintext="This is a message to everyone.",
   message_body_html="""<html><h1>Hello World!</h1><p>This is a message to all.</p></html>""", 
   email_recipients=['ghukill@mit.edu']
)
```

That is _very_ easy to test, without an `Report` objects needed.  But I take your point: do we _want_ to allow a non-standardized report like this to be possible.

If you require a `Report` object for sending a report, and I think I agree, then I'm still looking for that compelling case why `send_report()` is on the Workflow.  Why not have a method like `.send()` on the Report class?

If you go that route, and it's the CLI that finally calls something like,
```python
report.send(email_recipients.split(","))
```

then I think you could explain it like this:
- Workflows are _completely_ unaware of what a "Report"  is
- the CLI is the glue that runs a workflow and then performs reporting

But all just thought experiments!  If ya'll are happy with it as-is, I can be too.  They appeal to me equally.

dsc/report.py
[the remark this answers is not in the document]
jonavellecuerdo: See changes in commit https://github.com/MITLibraries/dspace-submission-composer/pull/121/commits/84652e9eb58376879f2a297c0b9629332e8801a7

dsc/report.py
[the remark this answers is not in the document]
jonavellecuerdo: See changes in commit https://github.com/MITLibraries/dspace-submission-composer/pull/121/commits/84652e9eb58376879f2a297c0b9629332e8801a7.

dsc/report.py
```diff
@@ -0,0 +1,190 @@
...
+from jinja2 import Environment, FileSystemLoader, select_autoescape
+
+if TYPE_CHECKING:
+    import dsc.workflows as workflows  # noqa: PLR0402
+
+
+class Report(ABC):
+    """A base report class from which other report classes are derived."""
+
+    template_name: str = "base"
```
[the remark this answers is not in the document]
jonavellecuerdo: Removed. See commit https://github.com/MITLibraries/dspace-submission-composer/pull/121/commits/84652e9eb58376879f2a297c0b9629332e8801a7.

dsc/cli.py
jonavellecuerdo: See https://github.com/MITLibraries/dspace-submission-composer/pull/121#discussion_r1946871007.

dsc/cli.py
jonavellecuerdo: Thank you for the input! 

* Re: "providing HTML being difficult" - Hmm, maybe I was thinking more down the line where we have a really beautiful HTML-formatted email message body. I can see myself wanting to capture this HTML in a file like a Jinja template. 🤔 
* I'm leaning towards "no" on non-standardized reports, at least at this stage 😅

dsc/reports/finalize.py
```diff
@@ -75,19 +68,27 @@ def create_attachments(self) -> list[tuple]:
...
+            for item in self.events.processed_items
+            if item["ingested"] == "success"
+        ]
+
+    def _write_ingested_items_csv(self, ingested_items: list[dict]) -> StringIO:
+        """Write ingested items to string buffer.
 
         This method creates a string buffer with the contents of a CSV
-        file describing successfully deposited items.
+        file describing successfully ingested items.
```
ehanson8, line 85: Good vocab switch to "ingested"

dsc/reports/finalize.py
```diff
@@ -18,9 +18,9 @@ class FinalizeReport(Report):
        and the number of encountered errors.
 
     2. A CSV file included as an attachment describing successfully deposited items,
-       which consists of the columns: 'item_identifier' and 'doi' (i.e., the 'ItemHandle'
-       from the DSS result message). Created only if any WorkflowEvents.processed_items
-       exist.
+       which consists of the columns: 'item_identifier' and 'dspace_handle' (i.e.,
+       the 'ItemHandle' from the DSS result message). Created only if
+       any items in Workflow.processed_items have ingested="success".
```
ghukill, line 23: Small wording changes, but I think reflecting good discussions last week.

dsc/reports/finalize.py
```diff
@@ -75,19 +68,27 @@ def create_attachments(self) -> list[tuple]:
             )
         return attachments
 
-    def _write_processed_items_csv(self) -> StringIO:
-        """Write processed items to string buffer.
+    def get_ingested_items(self) -> list[dict]:
```
ghukill, line 71: I dig it! Nice and explicit.  This basically codifies than an "ingested" item is a "processed" item where the SQS message indicates "success".

dsc/workflows/base/__init__.py
```diff
@@ -313,27 +313,35 @@ def process_sqs_queue(self) -> list[dict]:
         sqs_client = SQSClient(
             region=CONFIG.aws_region_name, queue_name=self.output_queue
         )
+
         items = []
         for sqs_message in sqs_client.receive():
             try:
-                item_identifier, message_body = sqs_client.process_result_message(
+                item_identifier, result_message = sqs_client.process_result_message(
```
ehanson8, line 320: This is technically just the message body that's returned by `process_result_message`, the result message contains the message attributes as well (where the only thing we care about currently is the `item_identifier` which we're already extracting)

ehanson8 (commented):
This is fantastic work and I would fully approve but there's one variable rename that I don't think is accurate unless I'm missing something. Otherwise, it looks great!

dsc/workflows/base/__init__.py
```diff
@@ -294,39 +313,56 @@ def process_sqs_queue(self) -> list[tuple[str, dict]]:
...
+            }
+            self.workflow_events.processed_items.append(item_data)
+            items.append(item_data)
+
+            if not item_data["ingested"]:
+                message = (
+                    f"Item '{item_identifier}' did not ingest successfully: {sqs_message}"
+                )
+                logger.info(message)
+                self.workflow_events.errors.append(message)
```
ghukill, line 343: After our discussions last week, I had to do a double/triple take here, but this feels right.

What I like about this flow:
- we _always_ put the processed item into the `WorkflowEvents.processed_items` list, whether ingested or not
- we save human-friendly message strings to `WorkflowEvents.errors`

Perhaps as reports grow in the future, we might want to lean even heavier into that and call it something like `WorkflowEvents.error_messages` or something, but maybe not required.

What we should be careful about -- and maybe this even suggests a docstring sometime -- is not treating `Workflow.WorkflowEvents` as data to feed into more work with the items.  My understanding is that it's 100% about reporting... they just so happen to be structured objects.

ghukill (approved):
Nice work!! Great discussions last week.  Feel good about where it landed here to move forward with.

PR #152, opened 2025-02-24
Add flexibility to SimpleCSV workflow

jonavellecuerdo:

This PR introduces changes that add flexibility to the SimpleCSV workflow, which were motivated by issues that were uncovered during test runs of two SCCS deposits (see [IN-1186](https://mitlibraries.atlassian.net/browse/IN-1186) for more details).

There are two commits in this PR:
* [updates to `S3Client.files_iter`](https://github.com/MITLibraries/dspace-submission-composer/pull/152/commits/04f731bf323b707912b5a105ce940067862cc3de)
* [updates to `SimpleCSV`](https://github.com/MITLibraries/dspace-submission-composer/pull/152/commits/a284e54cc85326e2b6369c13750998307a49f216)

I wrote most of the details in the commit messages, so please refer to them for more details!

The main change to `SimpleCSV` is in the `reconcile` methods. As discussed in the ticket, it became clear that retrieving item identifiers from bitstream filenames isn't possible _unless_ file naming conventions are strictly followed and/or regular expressions are used to parse item identifiers from filenames (doesn't seem ideal if file naming conventions vary across deposits as we'd have to potentially write new regex to handle each new variation). The **best** information that `reconcile` can provide is:
* what _item identifiers_ from the metadata CSV file matched to bitstreams (i.e., `<item_identifier> in <filename>`)
* what _filenames_ for bitstreams did not match
The changes in the second commit implement this change. 

1. In your terminal, navigate to the `dspace-submission-composer` repo and checking into `IN-1186-simplecsv-fix-item-identifiers`. 
2. Set AWS credentials for Dev.
3. Temporarily update `SCCS` workflow class attributes:
   ```text
   s3_bucket -> "wiley-files-dev-222053980223"
   output_queue -> "dss-wiley-output-dev"
   ```
4. Run `reconcile` command:
   ```shell
   pipenv run dsc --workflow-name sccs --batch-id sccs-batch-2025-02-09-mit-annual-reports reconcile                                    
   ```
   
   You should get the following output: 
```
2025-02-25 08:35:20,141 INFO dsc.cli.main(): Logger 'root' configured with level=INFO
2025-02-25 08:35:20,141 INFO dsc.cli.main(): No Sentry DSN found, exceptions will not be sent to Sentry
2025-02-25 08:35:20,141 INFO dsc.cli.main(): Running process
2025-02-25 08:35:20,142 INFO dsc.workflows.base.simple_csv.reconcile_bitstreams_and_metadata(): Reconciling bitstreams and metadata for batch 'sccs-batch-2025-02-09-mit-annual-reports'
2025-02-25 08:35:22,310 INFO dsc.workflows.base.simple_csv.reconcile_bitstreams_and_metadata(): Successfully reconciled bitstreams and metadata for all 106 item(s)
2025-02-25 08:35:22,312 INFO dsc.cli.post_main_group_subcommand(): Application exiting
2025-02-25 08:35:22,312 INFO dsc.cli.post_main_group_subcommand(): Total time elapsed: 0:00:02.171109
```

YES - Installs `pandas` 

YES - The column in the metadata CSV file that contains the item identifier
can take on any name as long as it is listed in the 'source_field_name'
for the 'item_identifier' entry in the metadata mapping JSON file.

- https://mitlibraries.atlassian.net/browse/IN-1186

### Developer

### Code Reviewer(s)

[IN-1186]: https://mitlibraries.atlassian.net/browse/IN-1186?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/dsc/workflows/base/simple_csv.py
+++ b/dsc/workflows/base/simple_csv.py
@@ -119,7 +119,7 @@ class SimpleCSV(Workflow):
             Item metadata.
         """
         with smart_open.open(
-            f"s3://{self.s3_bucket}/{self.batch_path}/{metadata_file}"
+            f"s3://{self.s3_bucket}/{self.batch_path}{metadata_file}"
         ) as csvfile:
             yield from csv.DictReader(csvfile)
 
--- a/dsc/workflows/base/simple_csv.py
+++ b/dsc/workflows/base/simple_csv.py
@@ -1,9 +1,11 @@
-import csv
+import itertools
 import json
 import logging
+from collections import defaultdict
 from collections.abc import Iterator
 from typing import Any
 
+import pandas as pd
 import smart_open
 
 from dsc.exceptions import ReconcileError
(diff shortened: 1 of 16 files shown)


dsc/workflows/base/simple_csv.py
```diff
@@ -152,5 +162,32 @@ def get_bitstream_s3_uris(self, item_identifier: str) -> list[str]:
         )
 
     def get_item_identifier(self, item_metadata: dict[str, Any]) -> str:
-        """Get 'item_identifier' from item metadata entry."""
-        return item_metadata["item_identifier"]
```
ghukill, line 156: Going to orient my comment here around this line removal, but it spans both this base `SimpleCSV` class and the more granular `SCCS` class.

I can appreciate the need for flexibility for any workflow derived from `SimpleCSV` that it might not be a column called `item_identifier`.  My question: instead of introducing this flexibility by making the base method lean on the workflow JSON mappings.... what if each workflow was required to define this method?

I had to slow down and carefully ready to make sure I understood this part:

> - If 'source_field_name' is a list (i.e., multiple field names accepted for 'item_identifier'), the method will return the value from the **first** mapped field that contains a non-null value.

I have a couple of interlated concerns:
  * do we risk confusion or hard to detect bugs if a CSV has _both_ fields defined here?  e.g. why did it use `hello-world.pdf` when we also have `hello-world` in column data?
  * I think there is some potentially unecessary misdirection here:
    * you start with the actual workflow class `SCSS`
    * you navigate to `SimpleCSV` to understand how it derives identifiers
    * this navigates you to `SCSS`'s JSON mapping file
    * there, you learn what fields are mapped to `[item_identifier, filename]`
    * you navigate _back_ to `SimpleCSV` and parse the logic about lists and first values encountered used

A method like `SCSS.get_item_identifier()`:

```python
def get_item_identifier(self, item_metadata: dict[str, Any]) -> str:
    for field in ['item_identifier','filename']:
        if item_identifier := item_metadata.get(field):
            return item_identifier
    raise ValueError("Failed to get item identifier from metadata")
```

The advantage I see here is that it's quite explicit a) which columns will be checked, and b) in what order, to extract an item identifier.

---

I'm going to change my mind before even submitting this comment.  I may see the appeal here.  Please pushback if this is _not_ the advantage of this approach.

By locating this in the workflow mapping, not only do we get some flexibility for extracting item identifiers, but we have already handled the creation of the actual DSpace metadata because it (re)uses the same mapping?

ghukill (commented):
Seeing some good changes in here, and not opposed to this approach of moving the item identifier flexibility into the workflow metadata JSON mapping.

Opted for a quick turnaround review with a fairly detailed question in a comment.

dsc/workflows/base/simple_csv.py
```diff
@@ -119,9 +125,13 @@ def item_metadata_iter(
...
         with smart_open.open(
-            f"s3://{self.s3_bucket}/{self.batch_path}/{metadata_file}"
+            f"s3://{self.s3_bucket}/{self.batch_path}{metadata_file}",
         ) as csvfile:
-            yield from csv.DictReader(csvfile)
+            metadata_df = pd.read_csv(csvfile, dtype="str")
+            metadata_df = metadata_df.dropna(how="all")
+
+            for _, row in metadata_df.iterrows():
+                yield row.to_dict()
```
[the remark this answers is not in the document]
jonavellecuerdo: Yes, definitely wanted to avoid that! I also encountered the tricky issue where the metadata CSV file I was working with had a TON of empty rows. The `drop.na` method was very handy!

dsc/workflows/base/simple_csv.py
jonavellecuerdo: Ah! Thank you for pointing this out! When you're in the weeds with the code, it can be hard to step back and think about the flow, so I really appreciate how you outlined "How do I know what source metadata fields were used to derive the 'item identifier'?"

I do think I prefer your suggestion to **define `SCCS.get_item_identifier`** for the following reason: 
> The 'item identifier' is **not** even mapped to any DSpace metadata fields (i.e., [it is not used by `Workflow.create_dspace_metadata`](https://github.com/MITLibraries/dspace-submission-composer/blob/IN-1186-simplecsv-fix-item-identifiers/dsc/workflows/base/__init__.py#L270)).

I noticed this while working on this ticket, but I was hesitant to suggest removing "item_identifier" from the metadata mapping JSON file for the sole reason of "to inform programmers what source metadata field is mapped to 'item_identifier"? _However,_ as you pointed out the journey to answering that question compels the programmer to bounce between `SimpleCSV` and `SCCS`.

---
In conclusion, I proposed to make the following changes
* Update `SimpleCSV.get_item_identifier` to simply: `return item_metadata["item_identifier"]`
* Define `SCCS.get_item_identifier` using your suggestion above
* Remove "item_identifier" from metadata mapping JSON files
   * I also like this idea as it further clarifies that the fields in the metadata mapping JSON files are the fields that end up in the generated DSpace metadata JSON file.

Let me know what you think!

ghukill: @jonavellecuerdo - glad the discussion is fruitful!  Makes sense to me.  Given that `item_identifier` does not end up in the DSpace metadata, and this change would further cement the role of the JSON mapping as creating DSpace metadata, I'm in favor of the changes too.

Furthermore, I think for this workflow and others going forward, a dedicated method like that is also easy to test to ensure it's picking up item identifiers as expected.

In short, proposal makes sense to me.  Even if refactored later, can't hurt to design towards known use cases now, avoid pre-mature abstraction.

jonavellecuerdo: 🍎 Fruitful indeed! Please see the latest commit https://github.com/MITLibraries/dspace-submission-composer/pull/152/commits/dc1944324589747df5a03b8385a00e914a7d45fa with the proposed updates.

ghukill: Looks good to me!

dsc/workflows/sccs.py
```diff
@@ -22,3 +24,17 @@ def s3_bucket(self) -> str:
...
         return "awaiting AWS infrastructure"
+
+    @staticmethod
+    def get_item_identifier(item_metadata: dict[str, Any]) -> str:
+        """Get 'item_identifier' from item metadata entry.
+
+        For SCCS deposits, this method expects at least one column
+        labeled "item_identifier" or "filename". If both of these
+        columns are present, the column labeled "item_identifier"
+        is selected.
```
ghukill, line 35: I think this is good.  While it suggests some loose-ish requirements on the input metadata we ask of stakeholders, it nonetheless explains how it's opinionated here.  Given the small footprint of the `SCSS` class, it really jumps out as a critical part of the class.

tests/test_workflow_simple_csv.py
```diff
@@ -20,51 +20,47 @@ def test_workflow_simple_csv_reconcile_bitstreams_and_metadata_success(
     ]
     simple_csv_workflow_instance.reconcile_bitstreams_and_metadata()
     assert (
-        "Successfully reconciled bitstreams and metadata for all items (n=1)"
+        "Successfully reconciled bitstreams and metadata for all 1 item(s)"
```
ghukill, line 23: 👍

ghukill (approved):
Looks good!  I think it's a nice adjustment given some testing on real world data.

PR #167, opened 2025-03-03
Add Infrastructure Build and Deploy Workflows

vab:

This PR adds the infrastructure deploy workflows to the this application, dsc. It adds new github automation, and updates the project's Makefile with the new workflows.

How this addresses that need:

    * Add dev build workflow
    * Add stage build workflow
    * Add prod promotion workflow
    * Update Makefile w/ new workflows
    * Add CODEOWNERS
    * Update README.md

Side effects of this change:

None

Changes to be committed:
      new file:   .github/CODEOWNERS
      new file:   .github/workflows/dev-build.yml
      new file:   .github/workflows/prod-promote.yml
      new file:   .github/workflows/stage-build.yml
      modified:   Makefile
      modified:   README.md

There will be no immediate effect. However, after the PR is merged with `main`, future code updates should allow the running of the deployment workflows.

NO

NO

https://mitlibraries.atlassian.net/browse/IN-1148

### Developer
- [X] All new ENV is documented in README
- [X] All new ENV has been added to staging and production environments
- [X] All related Jira tickets are linked in commit message(s)
- [X] Stakeholder approval has been confirmed (or is not needed)

### Code Reviewer(s)

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -1,0 +1,20 @@
+# CODEOWNERS file (from GitHub template at
+# https://docs.github.com/en/repositories/managing-your-repositorys-settings-and-features/customizing-your-repository/about-code-owners)
+# Each line is a file pattern followed by one or more owners.
+
+##############################################################################
+# These owners will be the default owners for everything in the repo. Unless a
+# later match takes precedence, @cabutlermit will be requested for review when
+# someone opens a pull request.This is commented out in favor of using a team
+# as the default (see below). It is left here as a comment to indicate the
+# primary expert for this code.
+# * @jonavellecuerdo
+
+# Teams can be specified as code owners as well. Teams should be identified in
+# the format @org/team-name. Teams must have explicit write access to the
+# repository.
+* @mitlibraries/dataeng
+
+# We set the senior engineer in the team as the owner of the CODEOWNERS file as
+# a layer of protection for unauthorized changes.
+/.github/CODEOWNERS @cabutlermit
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -4,10 +4,10 @@
 
 ##############################################################################
 # These owners will be the default owners for everything in the repo. Unless a
-# later match takes precedence, @cabutlermit will be requested for review when
-# someone opens a pull request.This is commented out in favor of using a team
-# as the default (see below). It is left here as a comment to indicate the
-# primary expert for this code.
+# later match takes precedence, @jonavellecuerdo will be requested for review
+# when someone opens a pull request.This is commented out in favor of using a
+# team as the default (see below). It is left here as a comment to indicate
+# the primary expert for this code.
 # * @jonavellecuerdo
 
 # Teams can be specified as code owners as well. Teams should be identified in
@@ -17,4 +17,4 @@
 
 # We set the senior engineer in the team as the owner of the CODEOWNERS file as
 # a layer of protection for unauthorized changes.
-/.github/CODEOWNERS @cabutlermit
+/.github/CODEOWNERS @ghukill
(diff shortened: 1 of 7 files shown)


.github/CODEOWNERS
```diff
@@ -0,0 +1,20 @@
...
+# * @jonavellecuerdo
+
+# Teams can be specified as code owners as well. Teams should be identified in
+# the format @org/team-name. Teams must have explicit write access to the
+# repository.
+* @mitlibraries/dataeng
+
+# We set the senior engineer in the team as the owner of the CODEOWNERS file as
+# a layer of protection for unauthorized changes.
+/.github/CODEOWNERS @cabutlermit
```
cabutlermit, line 20: This should be Graham (ghukill) not me, since this is a DataEng application.

Dockerfile
```diff
@@ -4,7 +4,7 @@ COPY . .
 
 RUN pip install --no-cache-dir --upgrade pip pipenv
 
-RUN apt-get update && apt-get upgrade -y && apt-get install -y git
+RUN apt-get update && apt-get dist-upgrade -y && apt-get autoremove -y && apt-get install -y git
```
cabutlermit, line 7: Since DataEng is building off the `python:3.12-slim` container, I don't think it's necessary to do the `dist-upgrade` here. I would leave any decisions regarding changes to the `Dockerfile` to the data engineers, especially @ghukill.

vab: I would recommend that they do a dist-upgrade and autoremove. The dist-upgrade will pick up changes related, through dependencies, to any packages that would be updated through an `apt-get upgrade`. This will help avoid ABI/API conflicts between packages and libraries. The autoremove command will uninstall packages that have been superseded during the upgrade or dist-upgrade process. This will save disk space (helping control our costs) and prevent either leaving insecure software installed in the container or a developer accidentally using software that has been replaced.

Performing our Dockerfile updates this way will bring us into line with the way AWS does their OS/Container updates. This would help prevent a situation where we could have containers that differ from our previous containers in installed software if we do a completely fresh build of a project or clone a project.

cabutlermit: This PR is supposed to be related just to the additional of the GitHub Actions workflows for publishing container images to AWS ECR, so any change to DataEng's `Dockerfile` feels out-of-scope for this PR.

Specific to this line in the `Dockerfile`, DataEng is using a pre-built Python container (`python:3.12-slim`), not a full Ubuntu distribution. So,the container is already highly optimized (see the `Dockerfile` that actually creates the source image [here](https://github.com/docker-library/python/blob/9c50e071f7445d04776485e28d208d39c25fbbaa/3.12/slim-bookworm/Dockerfile)).

It does make sense to me to do `apt-get upgrade` to catch any out-of-date packages, but an `apt-get dist-upgrade` goes beyond what is necessary for the Python application to actually run.

The dependencies that matter for the Python application are all handled by `pipenv install` line below (since DataEng is tracking all their Python dependencies in the `Pipfile` in their application code).

ghukill: Thanks for raising this up @vab , and the detailed reponse @cabutlermit.  I agree with Christopher's last three points about a specific image used, likely not needing to think/touch the underlying distribution, and that we spend quite a bit of time in our application code managing dependencies.

To my thinking -- and perhaps this is incorrect -- by tethering ourselves closely to the Docker image (and whatever updates it may receieve, knowing we get the updated form each CI build), we are effectively saying that our dependencies work with this particular image version more than `python 3.12` or `Ubuntu 22.04` (arbitrary examples) more generally.

vab: @ghukill I have made the requested change. Can you approve this PR for merge? It will unblock me on some tickets.  Thanks!

ghukill (approved):

PR #174, opened 2025-06-06
1319 create separate folder for dspace metadata

jonavellecuerdo:
By keeping the DSpace metadata JSON files created by DSC (via the submit command) in the same folder as the submission assets for a batch, these files may be accidentally included as “bitstreams” during reruns of the DSC workflow. Rather than excluding filenames that include the word “metadata” when retrieving bitstreams, DSC can be updated to exclude a metadata folder. 

A test execution of DSC step function in Dev1 with a `submit` payload using a sample of OpenCourseWare files: [3302ad0f-03e5-44d4-8f98-526854b492fa](https://us-east-1.console.aws.amazon.com/states/home?region=us-east-1#/v2/executions/details/arn:aws:states:us-east-1:222053980223:execution:dsc-process-dev:3302ad0f-03e5-44d4-8f98-526854b492fa).

1. Sign into the AWS Console.
2. On S3, navigate to [s3://dsc-dev-222053980223/opencourseware/opencourseware-2025-04-22-stepfunctiontest/dspace_metadata/](https://us-east-1.console.aws.amazon.com/s3/buckets/dsc-dev-222053980223?region=us-east-1&bucketType=general&prefix=opencourseware/opencourseware-2025-04-22-stepfunctiontest/dspace_metadata/&showversions=false)
3. Confirm that there are 5 DSpace metadata JSON files.

NO

YES - DSpace metadata JSON files will now be written to a subfolder `dspace_metadata/` within the batch folder on S3.

- https://mitlibraries.atlassian.net/browse/IN-1319

### Developer

### Code Reviewer(s)

### DIFF
--- a/dsc/workflows/base/__init__.py
+++ b/dsc/workflows/base/__init__.py
@@ -64,6 +64,7 @@ class Workflow(ABC):
         """
         self.batch_id = batch_id
         self.workflow_events = WorkflowEvents()
+        self.exclude_prefixes: list[str] = ["archived/", "dspace_metadata/"]
 
     @property
     @abstractmethod
--- a/dsc/workflows/base/simple_csv.py
+++ b/dsc/workflows/base/simple_csv.py
@@ -180,7 +180,7 @@ class SimpleCSV(Workflow):
                 bucket=self.s3_bucket,
                 prefix=self.batch_path,
                 item_identifier=item_identifier,
-                exclude_prefixes=["archived", "metadata.json"],
+                exclude_prefixes=self.exclude_prefixes,
             )
         )
 
--- a/dsc/workflows/opencourseware.py
+++ b/dsc/workflows/opencourseware.py
@@ -229,6 +229,6 @@ class OpenCourseWare(Workflow):
                 prefix=self.batch_path,
                 item_identifier=item_identifier,
                 file_type=".zip",
-                exclude_prefixes=["archived"],
+                exclude_prefixes=self.exclude_prefixes,
             )
         )


dsc/utilities/aws/s3.py
```diff
@@ -40,22 +38,33 @@ def archive_file_with_new_key(
         )
 
     def put_file(
-        self, file_content: str | bytes, bucket: str, key: str
-    ) -> PutObjectOutputTypeDef:
-        """Put a file in a specified S3 bucket with a specified key.
+        self,
+        bucket: str,
+        key: str,
+        file_content: str | bytes = b"",
```
ghukill, line 44: I'm curious, why might we have a default for an empty byte string?

dsc/workflows/base/__init__.py
```diff
@@ -151,13 +152,20 @@ def submit_items(self, collection_handle: str) -> list:
             "errors": 0,
         }
 
+        # create subfolder for DSpace metadata files in batch folder
+        metadata_folder = f"{self.batch_path}dspace_metadata/"
+        s3_client = S3Client()
+        s3_client.put_file(bucket=self.s3_bucket, key=metadata_folder)
```
ghukill, line 158: I have a question comment above about why `S3Client.put_file()` has an empty byte string default, but I think I'm understanding more given this usage.  Locating a comment here that has some tendrils.

Is the idea behind a zero-byte object to ensure that the "directory" is created in S3?  Is there a reason why we'd need a placeholder like this, if no objects that had that prefix were attempted or successfully created?

I suppose I might see a couple of options here, depending on the answer.

1- Leave `S3Client.put_file()` as-is, effectively doing what it "says on the tin" of putting a file in S3.  Also skip this work of creating an zero-byte object to establish the directory.  Instead, rely on metadata files actually written to establish that directory.

2- Continue writing a zero-byte object (perhaps there is good reason) but create a new method to do so, e.g. `S3Client.create_directory()`.  If `S3Client.put_file()` was unmodified, then `S3Client.create_directory()` could just _explicitly_ pass a `b""` argument for the file contents.  I think if I saw that method, I'd understand that for some reason -- and maybe the docstring could explain use cases -- why and how we are establishing zero-byte objects to represent a directory in S3.

ghukill (commented):
First things first, I think this is a great change!  I think it'll keep the workspace much more understandable for an item.

I had a couple of questions, mostly around why we need a zero-byte object to establish a directory, and the ramifications if that's _not_ needed.

dsc/workflows/base/__init__.py
ehanson8: I agree that much of this code may be unnecessary unless there's something I'm missing.

I think option 1 is the way to go. We can just create the "directory" with the existing `put_file()` call, I'll comment where the metadata file is created

dsc/item_submission.py
```diff
@@ -36,7 +36,11 @@ def upload_dspace_metadata(self, bucket: str, prefix: str) -> None:
         """
         s3_client = S3Client()
         metadata_s3_key = f"{prefix}{self.item_identifier}_metadata.json"
```
ehanson8, line 38: Could this just be updated to `metadata_s3_key = f"{prefix}/dspace_metadata/{self.item_identifier}_metadata.json"`?

Then I believe this code in `dsc/workflows/base/__init__.py` is unnecessary:

```
# create subfolder for DSpace metadata files in batch folder
metadata_folder = f"{self.batch_path}dspace_metadata/"
s3_client = S3Client()
s3_client.put_file(bucket=self.s3_bucket, key=metadata_folder)
```

We probably also need to update the docstring re: the `prefix` arg to include references to the `batch_path`, I think it was written before we had fully fleshed out the `batch_path` concept

ehanson8 (commented):
Made some comments on how this could be simpler but please push back if I'm missing something!

dsc/utilities/aws/s3.py
jonavellecuerdo: Resolving this comment and addressing in latter comment.

dsc/item_submission.py
jonavellecuerdo: Great suggestion! See [updates](https://github.com/MITLibraries/dspace-submission-composer/pull/175/files#diff-99c0807f75c8ef71eef8d7a50e3518ff669cfde4a65aa443245cf99658e15647R34-R45).

PR #181, opened 2025-07-02
IN-1317 Update submit CLI command to use DynamoDB

ehanson8:
Updates the `submit` CLI command to utilize DynamoDB. This required a method for writing `submit` results to DynamoDB and an `allow_submission` to skip sending a message when certain statuses are found in DynamoDB.

There were several items in the Jira ticket that did not happen as expected:

1. `last_submission_message` was an attribute in the DynamoDB schema, however, we only ever create the `message_attributes` and `message_body` for sending to the input queue. I would argue it is not worth assembling an "input message" just to record it in the table given that the `result_message` is far more important to record. Let me know if you disagree!
2. Given that `errors` are not associated with an `item_identifier` in `WorkflowEvents`, it's not possible to support the `status_details` attribute at this time. I'll make a note on @ghukill's [Jira ticket](https://mitlibraries.atlassian.net/browse/IN-1330) to consider this while working on this ticket.
3. While we discussed not using a `submit` retry threshold on Slack, I'm still recording `submit_attempts` for informational purposes.

Not possible yet

NO

NO

- https://mitlibraries.atlassian.net/browse/IN-1317

### Developer

### Code Reviewer(s)


dsc/workflows/base/__init__.py
```diff
@@ -425,6 +455,60 @@ def get_bitstream_s3_uris(self, item_identifier: str) -> list[str]:
             item_identifier: The identifier used for locating the item's bitstreams.
         """
 
+    @final
+    def allow_submission(self, item_identifier: str) -> bool:
+        """Verify that a submission message should be sent based on status in DynamoDB."""
+        item_submission_record = ItemSubmissionDB.get(
+            hash_key=self.batch_id, range_key=item_identifier
+        )
```
ghukill, line 463: Are we assuming that a row will exist here, created by the reconcile step?  

If for some reason it does not exist, e.g. reconciliation didn't happen, what should we do?

ghukill: Would it make sense to use `.get_or_create()` and just kind of ignore that reconciliation didn't happen?  or do we want to bail immediately if a row doesn't exist?

dsc/workflows/base/__init__.py
```diff
@@ -425,6 +455,60 @@ def get_bitstream_s3_uris(self, item_identifier: str) -> list[str]:
...
+        """Write submit results for each item submission to DynamoDB."""
+        item_submission_record = ItemSubmissionDB.get(
+            hash_key=self.batch_id, range_key=item_identifier
+        )
+        item_submission_record.update(
+            actions=[
+                ItemSubmissionDB.status.set(status),
+                ItemSubmissionDB.last_run_date.set(run_date),
+                ItemSubmissionDB.submit_attempts.set(
+                    ItemSubmissionDB.submit_attempts + 1
```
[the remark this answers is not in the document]
ehanson8: Excellent, thank you!

dsc/workflows/base/__init__.py
```diff
@@ -425,6 +455,60 @@ def get_bitstream_s3_uris(self, item_identifier: str) -> list[str]:
...
+        elif item_submission_record.status in [
+            None,
+            ItemSubmissionStatus.RECONCILE_FAILED,
+        ]:
+            logger.info(f"Record {log_details} not reconciled, skipping submission")
+
+        else:
+            allow_submission = True
+
+        return allow_submission
```
ghukill, line 489: At this stage in a project, I'm not sure that syntactical comments or suggestions are very helpful, but this felt like potentially a very good fit for a `match/case` statement.  Submitted for consideration!

```python
allow_submission = False

match item_submission_record.status:
    case ItemSubmissionStatus.INGEST_SUCCESS:
        logger.info(f"Record {log_details} already ingested, skipping submission")
    case ItemSubmissionStatus.SUBMIT_SUCCESS:
        logger.info(
            f"Record {log_details} already submitted, skipping submission"
        )
    case ItemSubmissionStatus.MAX_RETRIES_REACHED:
        logger.info(
            f"Record {log_details} max retries reached, skipping submission"
        )
    case None | ItemSubmissionStatus.RECONCILE_FAILED:
        logger.info(f"Record {log_details} not reconciled, skipping submission")
    case _:
        logger.debug(f"Record {log_details} allowed for submission")
        allow_submission = True

return allow_submission
```

ghukill (commented):
Sounds like changes are coming.  Submitting some initial comments at this time.

dsc/workflows/base/__init__.py
```diff
@@ -250,6 +255,11 @@ def submit_items(self, collection_handle: str) -> list:
             submission_summary["total"] += 1
             item_identifier = item_submission.item_identifier
 
+            # validate whether a message should be sent for this item submission
+            if not self.allow_submission(item_identifier):
+                continue
```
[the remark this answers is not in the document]
ehanson8: Good call!

dsc/workflows/base/__init__.py
ehanson8: I agree that a row shouldn't be created by `submit_items` given that it wouldn't have been successfully reconciled.

> * logging a warning that the item was not found in DynamoDB
> * including it in the "skipped" count
> * then `continue`

This all seems sound to me, I'll add it!

dsc/workflows/base/__init__.py
ehanson8: I like this! Updated!

README.md
```diff
@@ -52,6 +52,7 @@ SENTRY_DSN=### If set to a valid Sentry DSN, enables Sentry exception monitoring
 WORKSPACE=### Set to `dev` for local development, this will be set to `stage` and `prod` in those environments by Terraform.
 AWS_REGION_NAME=### Default AWS region.
 ITEM_TABLE_NAME=### The name of the table in DynamoDB used for tracking the state of an 'item' across DSC workflow executions.
+RETRY_THRESHOLD=### The maximum number of attempts to make when depositing an item submission into DSpace.
```
jonavellecuerdo, line 55: Hmm, maybe
> The maximum number of times an item submission can be retried for deposit into DSpace. Each submission's attempts are tracked in DynamoDB and incremented whenever a result message is processed for that submission.

dsc/config.py
```diff
@@ -37,6 +38,10 @@ def item_table_name(self) -> str:
             raise OSError("Env var 'ITEM_TABLE_NAME' must be defined")
         return value
 
+    @property
+    def retry_threshold(self) -> int:
+        return int(os.getenv("RETRY_THRESHOLD", "20"))
+
```
jonavellecuerdo, line 44: Can be simplified to pass an `int` to `os.getenv`:

```python
os.getenv("RETRY_THRESHOLD", 20)
```

tests/test_workflow.py
```diff
@@ -154,6 +244,89 @@ def test_base_workflow_validate_dspace_metadata_invalid_raises_exception(
         base_workflow_instance.validate_dspace_metadata({})
 
 
+def test_allow_submission_success(base_workflow_instance, mocked_item_submission_db):
```
jonavellecuerdo, line 247: Commenting here but applies to other test names. I'd update prefix to:

```text
test_base_workflow...
```

jonavellecuerdo (changes requested):
Looking good! Thank you for addressing the previous review. Here is a smaller batch of minor change requests. Thanks!

dsc/config.py
ghukill: @jonavellecuerdo - not 100% sure about that?  If we want to be gauranteed an integer coming out of this, my working assumption has always been that env var values always start as strings.

I think wrapping with `int()` and defaulting to a string `"20"` is a pretty standard way of doing it?

ghukill (approved):
Approved, with some accompanying thoughts.

While looking at the `Workflow.submit_items()` method, something that immediately jumps out is the following codeblock repeating:

```python
item_submission_record.update(
    actions=[
        ItemSubmissionDB.status.set(ItemSubmissionStatus.SUBMIT_FAILED),
        ItemSubmissionDB.status_details.set(str(exception)),
        ItemSubmissionDB.last_run_date.set(self.run_date),
        ItemSubmissionDB.submit_attempts.add(1),
    ]
)
```

A naive suggestion would be to refactor to a method and have this method call that.  But I think this excerbates _another_ issue of the `Workflow` class  growing in size and complexity.  Reconcilation, now submit, and finalize _will_, are all introducing additional similar but not identical methods as they perform some kind of "loop through things, do things, save things to DB, and report things".  

My fear is that  focusing on any given subset -- e.g. `submit_items()` here -- would miss an opportunity to think more globally about this.  For example, what if `Workflow` was _composed_ of multiple other more focused classes like `ReconcileItems`, `SubmitItems`, and `FinalizeItems` a la:

```python
class Workflow():
    def __init__(self):
        self.reconcile = Reconcile()
        self.submit = Submit()
        self.finalize = Finalize()
```

where then each of those classes might have something like a `.run()` method  that would perform their work.  Perhaps while doing _that_ refactor we notice they all keep doing similar DyanmoDB updating of items.  It would stand to reason either a) a standalone function they could call to update items or b) moving that to more methods on the `ItemSubmissionDB` class.  

To be clear: I think this code as-is will function!  And I think at this time that's ultimately more valuable to continue exposing contours of this project we may not yet understand.  But similar to the Dynamo + reconcile PR, I think this PR is hinting that something more structural at the `Workflow` and `ItemSubmissionDB` level isn't _quite_ right and it's rearing up in these methods, making them _impossible_ to really nail.

I say all this as the most removed person on this project, and without a clear vision or proposal of what that reworking looks like.  But my gut feeling is that we should continue plowing forward with the functionality like this PR adds, e.g. moving onto finalize, and then pause for a moment and look for some fairly deep structural refactorings that might make all the pieces snap together a bit cleaner.  

What I _do_ like about this PR, and something I had written and deleted multiple comments about, is the very procedural nature of the code.  Because we have _not_ yet introduced  too many methods which increase misdirection, I think we're setup well for some refactoring.

**TL/DR:** Approved with the belief we keep moving forward towards end-to-end functionality, then pause and take stock for a potentially fairly substantial refactor of _working_ functionality.

README.md
ehanson8: That's better, thank you!

dsc/workflows/base/__init__.py
```diff
@@ -250,6 +253,25 @@ def submit_items(self, collection_handle: str) -> list:
...
+            # retrieve record from DynamoDB
+            try:
+                item_submission_record = ItemSubmissionDB.get(
+                    hash_key=self.batch_id, range_key=item_identifier
+                )
+            except DoesNotExist:
+                logger.warning(
+                    f"DynamoDB row not found for '{item_identifier}', verify that it has "
+                    "been reconciled."
+                )
```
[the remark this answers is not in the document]
ehanson8: Done, this is a much better approach, thank you for suggesting!

dsc/config.py
jonavellecuerdo: I retract my request!

jonavellecuerdo (approved):
Approved! Thank you for the additional cleanup. ✨

PR #189, opened 2025-07-29
In 1356 update submit

ehanson8:
Further updates to take advantage of new `ItemSubmission` functionality in the `submit` command.

Not possible at this stage

NO

NO

- https://mitlibraries.atlassian.net/browse/IN-1356

### Developer

### Code Reviewer(s)

### DIFF
--- a/dsc/workflows/base/__init__.py
+++ b/dsc/workflows/base/__init__.py
@@ -261,7 +261,7 @@ class Workflow(ABC):
         }
 
         items = []
-        for item_submission_record in ItemSubmissionDB.get_batch_items(self.batch_id):
+        for item_submission_record in ItemSubmissionDB.query(self.batch_id):
 
             # instantiate ItemSubmission instance from DB record
             item_submission = ItemSubmission.from_db(item_submission_record)
--- a/dsc/db/models.py
+++ b/dsc/db/models.py
@@ -195,15 +195,6 @@ class ItemSubmissionDB(Model):
 
         return item
 
-    @classmethod
-    def get_batch_items(cls, batch_id: str) -> list["ItemSubmissionDB"]:
-        """Get all items in a batch.
-
-        Args:
-            batch_id: The unique identifier for the workflow run.
-        """
-        return list(cls.query(batch_id))
-
     def to_dict(self, *attributes: str) -> dict:
         """Create dict representing an item submission.
 
--- a/tests/test_db.py
+++ b/tests/test_db.py
@@ -61,18 +61,3 @@ def test_db_itemsubmission_get_or_create_success(mocked_item_submission_db):
     assert fetched_item.item_identifier == "123"
     assert fetched_item.batch_id == "batch-aaa"
     assert fetched_item.workflow_name == "workflow"
-
-
-def test_db_itemsubmission_get_batch_items_success(mocked_item_submission_db):
-    ItemSubmissionDB.create(
-        batch_id="batch-aaa", item_identifier="123", workflow_name="workflow"
-    )
-    ItemSubmissionDB.create(
-        batch_id="batch-bbb", item_identifier="1234", workflow_name="workflow"
-    )
-
-    items = ItemSubmissionDB.get_batch_items(batch_id="batch-aaa")
-
-    assert len(items) == 1
-    assert items[0].item_identifier == "123"
-    assert all(item.batch_id == "batch-aaa" for item in items)


tests/test_workflow_base.py
```diff
@@ -86,34 +86,6 @@ def test_base_workflow_submit_items_success(
     assert json.dumps(expected_submission_summary) in caplog.text
 
 
-def test_base_workflow_submit_items_missing_row_raises_warning(
```
ehanson8, line 91: I don't believe this scenario is still possible as we're instantiating from `ItemSubmissionDB.get_batch_items` but please correct me if I'm wrong!

dsc/workflows/base/__init__.py
```diff
@@ -313,30 +308,25 @@ def item_submissions_iter(self) -> Iterator[ItemSubmission]:
...
-                    "Record "
-                    f"{ITEM_SUBMISSION_LOG_STR.format(batch_id=self.batch_id,
-                                      item_identifier=item_identifier)}"
-                    "not found. Verify that it been reconciled."
-                )
-                self.submission_summary["skipped"] += 1
-                continue
-
-            logger.info(f"Preparing submission for item: {item_identifier}")
+            logger.info(
```
ghukill, line 318: It's minor, but I'd propose `DEBUG` logging here.  Overall, I think we should lean into `DEBUG` when inside a loop unless we have a very good reason not too.

Also, a much fuzzier heuristic might be `DEBUG` can help understand when something goes wrong.  If the item loaded successfully, awesome, maybe we don't need to know that in the logs.

dsc/workflows/base/__init__.py
```diff
@@ -313,30 +308,25 @@ def item_submissions_iter(self) -> Iterator[ItemSubmission]:
 
         MUST NOT be overridden by workflow subclasses.
         """
-        for item_metadata in self.item_metadata_iter():
```
ghukill, line 316: _I can't comment on the method signature itself, so going to focus this comment here._

Curious your thoughts on pulling the whole dspace metadata retrieval + generation _outside_ of this method?

On my first read through, I found myself stumbling on what this method is doing.  I had expected something like:

```python
for item_db in ItemSubmissionDB.query():
   yield ItemSubmission.from_db(item_db)
```

I could imagine it doing some bookkeeping work, e.g. stuff like `self.submission_summary["total"] += 1`, but I was quite surprised to see it retrieving and generating the final DSpace metadata.

Instead, I would expect to see that in the main `Workflow.submit_items()` method as part of the primary work.  This might come with a couple of related tendrils:

1. Maybe the method `ItemSubmission.upload_dspace_metadata()` should expect the metadats as a passed argument, instead of looking to `self`.   This would allow generating that metadata as part of the main `submit_items()` flow, then passing it there for upload.  This might further suggest _removing_ `ItemSubmission.dspace_metadata` as a class attribute; we can just generate + pass it when needed.  

2. Similar story with `get_bitstream_s3_uris()` which feel like part of the work of `Workflow.submit_items()`.

3. This might _further_ suggest that even `Workflow.item_submissions_iter()` is not needed?  If it's as simple as above, then maybe it could move back to the `Workflow.submit_items()` method and remove some of that misdirection.

Here is what I kind of expected to see in `Workflow.submit_items()`, not that it's correct, but just what I found surprising on this PR:

```python
def submit_items(self, ...):
    
    for item_db in ItemSubmissionDB.get_batch():
        
        # hydrate from database
        item_submission = self.from_db(item_db)
        
        # check if ready for submission
        if not item_submission.ready_to_submit():
            self.submission_summary["skipped"] += 1
            continue
        
        # generate dspace metadata
        dspace_metadata = self.create_dspace_metadata(...)
        
        # upload metadata
        item_submission.upload_dspace_metadata(dspace_metadata)
        
        # get bitstream URIs
        bitstream_s3_uris = self.get_bitstream_s3_uris()
        
        # publish SQS message
        item_submission.send_submission_message(bitstream_s3_uris)
        
        # bookkeeping
        item_submission.status = ItemSubmissionStatus.SUBMIT_SUCCESS
        item_submission.save()
```

I'll admit, when sketching it like this, I begin to wonder why `upload_dspace_metadata()` and `send_submission_message()` are methods on `ItemSubmission`.  I know we had talked about the agency of an `ItemSubmission` class instance, but it feels like it should work in harmony with the `Workflow` instance.

A bit of an example here.  And, **fully acknowledging this passed code review last time, but didn't jump out like it is now**.

The method `ItemSubmission.send_submission_message` has this block:

```python
        except ClientError as exception:
            logger.error(  # noqa: TRY400
                f"Failed to send submission message for item: {self.item_identifier}. "
                f"{exception}"
            )

            self.status = ItemSubmissionStatus.SUBMIT_FAILED
            self.status_details = str(exception)
            self.submit_attempts += 1
            self.update_db()
            raise
```

While it makes sense that an `ItemSubmission` can update itself in the DB, it doesn't feel quite right that it would do so when one of its methods fail.

Imagine this raises a custom exception like `ItemSubmitError("could not send message to SQS")`.  Then back in the `Workflow.submit_items()` loop you could have error handling like this:

```python
try:
    # lots of things...
except ItemSubmitError as exc:
    item_submission.ItemSubmissionStatus.SUBMIT_FAILED
    item_submission.update_db()
```

I haven't checked, but I imagine multiple methods on `ItemSubmission` are doing something similar?  If they _all_ just raise exceptions then you can let the "orchestrator" of `Workflow.submit_items()` decide on how to catch them and then update the DB.

ghukill (commented):
Despite my lengthy single comment, I can see the pieces here.  Before going further in my review, wanted to pause for discussion on that comment.

It feels like the functionality is here but I'm finding myself tripping on the responsibilities and flow of `Workflow` and `ItemSubmission`.  Happy to huddle and discuss if the comment alone is not helpful.

dsc/workflows/base/__init__.py
ghukill: On further reflection, I can see why maybe these didn't pop in the last PR.  The refactoring of `ItemSubmission` and this `submit_items()` workflow are just _really_ intertwined.  

It felt like it made some sense last time that `ItemSubmission` would have all that agency, but seeing it now in the loop/work of `Workflow.submit_items()`, it feels like that **orchestrating** method should be doing more of that orchestration.

ehanson8: Digesting this and going to take another pass, @jonavellecuerdo you might hold off if you haven't started yet

ehanson8: ✔️ I've done away with `Workflow.item_submission_iter` already and reordered things a bit including shifting DSpace metadata creation out of the instantiation loop which is a relic of the earlier flow, I'll push a commit soon. It feels like a toss up between `Workflow` and `ItemSubmission` on some of this functionality given that attributes are needed from both but it'll be easier to discuss in the context of a new commit

dsc/workflows/base/__init__.py
```diff
@@ -296,40 +313,17 @@ def submit_items(self, collection_handle: str) -> list:
                 self.workflow_events.errors.append(str(exception))
                 self.submission_summary["errors"] += 1
 
+                item_submission.status = ItemSubmissionStatus.SUBMIT_FAILED
+                item_submission.status_details = str(exception)
+                item_submission.submit_attempts += 1
+                item_submission.update_db()
+
```
ehanson8, line 320: This block is pointing to the tension between tracking in DynamoDB vs. workflow events but leaving this here for now

dsc/workflows/base/__init__.py
```diff
@@ -296,40 +313,17 @@ def submit_items(self, collection_handle: str) -> list:
                 self.workflow_events.errors.append(str(exception))
                 self.submission_summary["errors"] += 1
 
+                item_submission.status = ItemSubmissionStatus.SUBMIT_FAILED
+                item_submission.status_details = str(exception)
+                item_submission.submit_attempts += 1
+                item_submission.update_db()
```
ghukill, line 319: This feels good here.  Checked the `ItemSubmission` class and not seeing any surprising side effects, just trying to do things and raising errors where needed.

ghukill (approved):
Approved!    Thanks for the discussions on this PR, and outside via huddles.

As mentioned in a recent discussion, it seems possible we might touch `ItemSubmission` methods like `from_db()`, and maybe this `Workflow.submit_items()` will get a light touch going forward, but the overall order and placement of things is feeling good.

Wanted to note one thing that feels a little odd, but unsure of a different approach.

The `Workflow.submit_items()` method it kicks off with this:

```python
batch_metadata = {
    self.get_item_identifier(item_metadata): item_metadata
    for item_metadata in self.item_metadata_iter()
}
```

I understand what it does, given that we loop through DB instances, we still need to parse this out so we can pass to `item_submission.prepare_dspace_metadata()`.

But it feels _really_ important for how little fanfare it gets there?  Like... should we parse this out and store it on the Workflow instance during initialization?   Probably not, as we don't need for finalize.  But I do wonder if there should be a method like `load_item_metadata()` or something that _uses_ the iterator method that each Workflow has, but it actually runs it and saves to `self` for usage.  Then it could have additional error reporting if something goes wrong, as I imagine this isn't the only place we'd use that method.

Again: not a requested change now, just noting.  

Otherwise, **nice work**!  Big changes, good forward progress.

ehanson8:
> ```python
> batch_metadata = {
>     self.get_item_identifier(item_metadata): item_metadata
>     for item_metadata in self.item_metadata_iter()
> }
> ```

Yes, I was realizing this could serve `reconcile` as well during the conversation earlier, @jonavellecuerdo I'm happy to add a method for this or you could tackle it in your PR, I'm fine either way! Should be simple enough to implement

dsc/workflows/base/__init__.py
```diff
@@ -260,19 +255,36 @@ def submit_items(self, collection_handle: str) -> list:
...
         )
 
+        batch_metadata = {
+            self.get_item_identifier(item_metadata): item_metadata
+            for item_metadata in self.item_metadata_iter()
+        }
+
         items = []
-        for item_submission in self.item_submissions_iter():
+        for item_submission_record in ItemSubmissionDB.get_batch_items(self.batch_id):
```
jonavellecuerdo, line 264: Hmm, I think this was briefly mentioned in the first part of [this comment](https://github.com/MITLibraries/dspace-submission-composer/pull/189#discussion_r2240783951), but it doesn't seem like [`ItemSubmissionDB.get_batch_items()` ](https://github.com/MITLibraries/dspace-submission-composer/blob/IN-1356-update-submit/dsc/db/models.py#L198-L205) does a lot beyond materializing all the elements in the generator. I propose:

a. Just using `ItemSubmissionDB.query` or
b. Update `ItemSubmissionDB.get_batch_items` to simply:
```python
return cls.query(batch_id)
```

I understand that having a `get_batch_items` method allows the method call to be more detailed explicit, but again, it really doesn't seem to be doing much (so perhaps **b** is a good compromise)? 🤔

PR #191, opened 2025-08-05
In 1357 finalize refactor

ehanson8:
Further updates to take advantage of new `ItemSubmission` functionality in the `finalize` command.

Not possible at this stage

NO

NO

- https://mitlibraries.atlassian.net/browse/IN-1357

### Developer

### Code Reviewer(s)

### DIFF
--- a/dsc/workflows/base/__init__.py
+++ b/dsc/workflows/base/__init__.py
@@ -419,7 +419,7 @@ class Workflow(ABC):
 
             result_message = result_message_map.get(item_submission.item_identifier)
 
-            # set status to ingest_unknown if no message exists with item_identifier
+            # skip item submission if result message is not found
             if not result_message:
                 continue
 


dsc/item_submission.py
```diff
@@ -146,6 +146,8 @@ def ready_to_submit(self) -> bool:
         """Check if the item submission is ready to be submitted."""
         ready_to_submit = False
 
+        self._exceeded_retry_threshold()
```
ehanson8, line 149: Realized the app was not checking the `retry_threshold` so adding this

dsc/workflows/base/__init__.py
ehanson8, line 1: Given that `process_result_messages` didn't seem as necessary as it did before (but feel free to disagree!), I removed it as a separate commit to avoid an awful diff in the next commit

dsc/workflows/base/__init__.py
```diff
@@ -315,6 +317,85 @@ def finalize_items(self) -> None:
...
+        )
+        result_message_map = {}
+        for message in sqs_client.receive():
+            try:
+                self.validate_result_message(message)
+                item_identifier = message["MessageAttributes"]["PackageID"]["StringValue"]
+                result_message_map[item_identifier] = message
+            except InvalidSQSMessageError:
+                logger.exception(f"Failure parsing message '{message["MessageId"]}'")
+                continue
```
ehanson8, line 345: Given the updated flow, invalid messages are not added to `result_message_map` which effectively triggers the same `INGEST_UNKNOWN` status via lines 363-368 rather in the previous `except` block for an individual message

ehanson8: Thinking about this more, the exception will no longer be recorded in the `status_details` attribute which may or may not be a problem. But I do think invalid result messages will be exceedingly rare, I *believe* DSS and SQS will bounce them well before this point, but we should still have checks in case they don't

dsc/workflows/base/__init__.py
```diff
@@ -371,119 +327,108 @@ def process_result_messages(self) -> None:
...
-                # delete message from the queue
-                sqs_client.delete(
-                    receipt_handle=receipt_handle,
-                    message_id=message_id,
-                )
+                self.validate_result_message(message)
+                item_identifier = message["MessageAttributes"]["PackageID"]["StringValue"]
+                result_message_map[item_identifier] = message
+            except InvalidSQSMessageError:
+                logger.exception(f"Failure parsing message '{message["MessageId"]}'")
```
ghukill, line 344: Feeling like maybe this could be a sub-method?

Would be great in this main `finalize()` method to have something like:

```python
result_message_map = self.retrieve_and_parse_result_messages()
```

dsc/workflows/base/__init__.py
```diff
@@ -371,119 +327,108 @@ def process_result_messages(self) -> None:
...
+                item_submission.upsert_db()
+                logger.info(f"Unable to determine ingest status for record {log_str}")
+                continue
 
-                new_status = ItemSubmissionStatus.INGEST_UNKNOWN
-
-                item_submission_record.update(
-                    actions=[ItemSubmissionDB.status_details.set(str(exception))]
+            # update item submission status based on ingest result
+            result_message_body = json.loads(result_message["Body"])
```
ghukill, line 371: Could we do this JSON loading during the result message parsing, e.g. as part of the proposed `retrieve_and_parse_result_messages()` sub-method?

dsc/workflows/base/__init__.py
ghukill: Circling back to this comment after going a little further in code, and I think a `@dataclass` could potentially be a nice fit to represent and encapsulate a result message.

It could provide a few things:
- factory method to parse a "raw" message and return an instance of itself
- internal methods for parsing the JSON of the message Body
- dot notation for common  things like item identifier, `Body.ResultType`, etc.
- etc.

It _might_ even be worth exploring if the `SQSClient()` -- which is a client opinionated for this application -- could yield instances of those.  But that might be a step too far if we just want to keep that client 100% pure  SQS in capabilities.

dsc/workflows/base/__init__.py
ghukill: Update: see above about possible `ResultMessage` `@dataclass`.

dsc/workflows/base/__init__.py
```diff
@@ -371,119 +327,108 @@ def process_result_messages(self) -> None:
...
+                receipt_handle=result_message["ReceiptHandle"],
+                message_id=result_message["MessageId"],
             )
-            logger.info(f"Status updated: '{current_status}' -> '{new_status}'")
 
-            # delete message from the queue
-            sqs_client.delete(
-                receipt_handle=receipt_handle,
-                message_id=message_id,
+        self.workflow_specific_processing()
```
ghukill, line 399: I feel like this was probably just ported from previously reviewed code, but it feels potentially too big / too important to be a single, undecorated line.

I see a couple of options:

1. Move this call back to the CLI.  We run a "normal" finalize process, then the CLI context calls `workflow.workflow_specific_processing()`.  Looking at in that light, it almost feels like it could/should  be renamed to something like `workflow.post_finalize_processing()`.  

2. Minor, but move to the very last line of the `finalize()`.  

Both of these assume that sending workflow events could happen before this workflow-specific work.  Is that true?  do we have a concrete use case for this yet?  is it more accurate to call it `post_finalize_processing()`?  or should we think of it as _part of_ `finalize`?

Maybe no changes are needed, but this line does jump out at me.

ghukill (commented):
Left a few questions/comments, but overall looking _really_ good!  Nice work.

I found when reading through it, the overall flow, comments, everything made sense.

I  think more encapsulation of the result message parsing could be helpful, and I'm curious about the workflow-specific processing.

dsc/workflows/base/__init__.py
ghukill: Ah, just seeing this now.  Curious your thoughts on the dedicated sub-method and/or `ResultMessage` dataclass I floated.

dsc/workflows/base/__init__.py
ehanson8: I like the `@dataclass` approach, let me implement that and see how everything looks afterward, @jonavellecuerdo you might hold off until I push those changes

dsc/workflows/base/__init__.py
ghukill: Cool cool!  Makes sense.  For now, if nothing else, maybe just a comment above that line?  The method is really well named, but I still worry about it getting lost in the shuffle via a pure lines of code numbers game.

The flipside, I think it could stand on its own if the code was shaped more like:

```python
result_message_map = self.retrieve_and_parse_result_messages()
for item_submission in ItemSubmission.get_batch(batch_id):
    self._finalize_item(item_submission)  #<----- new submethod that captures 99% of the work
self.workflow_specific_processing()
```

I am *not* proposing that change, just trying to sketch where I think `self.workflow_specific_processing()` would look more normal in scale.

ehanson8: Added a comment!

dsc/workflows/base/__init__.py
ehanson8: Added a `@dataclass`, let me know what you both think!

dsc/workflows/base/__init__.py
```diff
@@ -357,45 +415,40 @@ def finalize_items(self) -> None:
 
             item_submission.ingest_attempts += 1
 
-            result_message = result_message_map.get(item_submission.item_identifier)
+            result_message = result_message_map.get(item_submission.item_identifier)  # type: ignore[assignment]
```
ehanson8, line 418: I could not figure out how stop `mypy` from complaining about this line without an invasive and seemingly duplicative check (I assumed it would be covered by line 421 but I guess not) so I added an `#type: ignore`

dsc/workflows/base/__init__.py
```diff
@@ -371,165 +386,88 @@ def process_result_messages(self) -> None:
...
-                    message_id=message_id,
-                )
+                result_message = DSSResultMessage.from_result_message(message)
+                result_message_map[result_message.item_identifier] = result_message
+            except InvalidSQSMessageError:
+                logger.exception(f"Failure parsing message '{message}'")
                 continue
 
-            item_identifier = message_attributes["PackageID"]["StringValue"]
+        sqs_results_summary["received_messages"] = len(result_message_map)
```
ghukill, line 405: Nice work here.  It feels easier to reason about without the grain of the message itself, focusing here more on the consumption of the SQS queue.

Do you think it's worth taking this a step further and encapsulating _this_ in a sub-method?  Now that the dataclass `ResultMessage` handles the parsing of an individual message, we could still take some cognitive load out of this `finalize()` method by refactoring the SQS consumption + parsing out. 

I'm okay either way.  In a weird way I think this SQS consumption _is_ the work of this `finalize()` method.  The flipside argument might be keeping this here, but refactoring the `if result_message.result_type == X` logic (roughly lines 428-444) into a sub-method that updates the `ItemSubmission` instance based on that logic, then finshing up here with `item_submission.upsert_db()` and `sqs_client.delete(...)`.

Not formally requesting changes, but observing this method still reads long; feels like some of this work could be farmed out to sub-methods keeping this high level, entrypoint `finalize()` method just farming out work as needed.

dsc/workflows/base/__init__.py
```diff
@@ -32,6 +33,65 @@
 )
 
 
+@dataclass
+class DSSResultMessage:
```
ghukill, line 37: Love it!

ghukill (approved):
Submitting an approval as-is, looks great.  With an optional comment around further sub-methods.  Happy to re-review if you go that route.

Nice work!

dsc/workflows/base/__init__.py
ehanson8: I went back and forth on this, I do think there's some symmetry with `submit_items` re: the abstraction-level (but maybe both are in need of a sub-methods!). @jonavellecuerdo Your thoughts? I'm fine to do it either way

ghukill: Taking another look at `submit_items`, it's got these sub-methods which I think do the majority of its "work":
- `prepare_dspace_metadata`
- `get_bitstream_s3_uris`
- `send_submission_message`

But, `finalize_items` kind of has less work to do?  It's more about parsing and recording.  So yeah, I'm kind of on the fence too.  Remain good with it as-is, where the _important_ part was shifting things to use `ItemSubmission` and `.get_batch()`.  Nothing stopping anyone from another pass near the end when maybe the relative size of reconcile, submit, and finalize would be more obvious.

dsc/workflows/base/__init__.py
```diff
@@ -371,165 +386,88 @@ def process_result_messages(self) -> None:
...
-                logger.error(  # noqa: TRY400
-                    f"Failed to parse 'Body' from {message_id}:{exception}"
-                )
+            # set status to ingest_unknown if no message exists with item_identifier
+            if not isinstance(result_message, DSSResultMessage):
+                item_submission.status = ItemSubmissionStatus.INGEST_UNKNOWN
                 sqs_results_summary["ingest_unknown"] += 1
+                item_submission.upsert_db()
+                logger.info(f"Unable to determine ingest status for record {log_str}")
+                continue
```
jonavellecuerdo, line 426: I spent some thinking over the effect of this update 🤔 and curious whether you, @ehanson8, and @ghukill arrived at the same understanding:

> For each item submission in the batch, if there wasn't a result message, the status will get updates to `ItemSubmissionStatus.INGEST_UNKNOWN. 

1. What happens if DSS was only run for a _subset_ of items in the batch (i.e., `submit` was run for a _subset_ of items, either by error or intentionally)?
   * Should the status remain the same?
   * Should `item_submission.ingest_attempts` be incremented?
      **Note:** Doesn't feel right to increment `ingest_attempts` if there was no result message in the queue for the item submission (i.e., `submit` + `dss` was not run for the item submission).

I think the reason why we didn't have this issue with [the previous version of the code](https://github.com/MITLibraries/dspace-submission-composer/blob/main/dsc/workflows/base/__init__.py#L378-L484) is because all DynamoDB updates were encapsulated within the for loop retrieving messages from the output queue. 

Please let me know if I'm misunderstanding -- curious to get both your thoughts!

dsc/workflows/base/__init__.py
jonavellecuerdo: I like the idea of moving it to the bottom of `Workflow.finalize_items`.

jonavellecuerdo (changes requested):
@ehanson8 Looking good, but I think there is an unintended side effect of updating statuses of `ItemSubmission` objects in the DynamoDB table even when there aren't any associated result messages received from the queue (more details in comment below)!

dsc/workflows/base/__init__.py
ghukill: Agreed, that seems like a nice catch of a possible edge case.

Supposing we did _nothing_ if we hit this block:

```python
if not isinstance(result_message, DSSResultMessage):
    continue # <------------------
```

Would that be problematic?  Wouldn't the previous status be whatever `submit_items` had set?

ehanson8: @jonavellecuerdo You are correct, great catch again and I think @ghukill's suggestion would work! I will also change the `isinstance` check to `if not result_message`, that was part of me fighting with `mypy`. @jonavellecuerdo Do you have any concern about @ghukill 's suggestion?

jonavellecuerdo: No concerns. I agree that @ghukill 's suggestion will work! Can you also move the code incrementing `ingest_attempts` to _after_ any `continue` statements:

```python
if item_submission.status == ItemSubmissionStatus.INGEST_SUCCESS:
   logger.debug(f"Record {log_str} already ingested, skipping")
   continue

result_message = result_message_map.get(item_submission.item_identifier)  # type: ignore[assignment]

if not result_message: 
   continue

item_submission.ingest_attempts += 1
```

dsc/workflows/base/__init__.py
```diff
@@ -415,14 +417,10 @@ def finalize_items(self) -> None:
...
-            result_message = result_message_map.get(item_submission.item_identifier)  # type: ignore[assignment]
+            result_message = result_message_map.get(item_submission.item_identifier)
 
             # set status to ingest_unknown if no message exists with item_identifier
-            if not isinstance(result_message, DSSResultMessage):
-                item_submission.status = ItemSubmissionStatus.INGEST_UNKNOWN
-                sqs_results_summary["ingest_unknown"] += 1
-                item_submission.upsert_db()
-                logger.info(f"Unable to determine ingest status for record {log_str}")
+            if not result_message:
```
jonavellecuerdo, line 423: Comment needs update!

ehanson8: Oops, fixed!

jonavellecuerdo (approved):

PR #192, opened 2025-08-13
Implement metadata transformer for OpenCourseWare workflow

jonavellecuerdo:

This PR proposes the use of a new "metadata transformer" class for the `OpenCourseWare` workflow. As the ticket name indicates, this was originally expected to be a simple update of the metadata mapping JSON file, but the [updated metadata mapping](https://mitlibraries.atlassian.net/browse/IN-1394?focusedCommentId=158864) proved to be more complex. 

NO

NO

- https://mitlibraries.atlassian.net/browse/IN-1394

### Developer

### Code Reviewer(s)

### DIFF
--- a/dsc/workflows/opencourseware.py
+++ b/dsc/workflows/opencourseware.py
@@ -1,7 +1,8 @@
+import inspect
 import json
 import logging
 import zipfile
-from collections.abc import Iterator
+from collections.abc import Iterable, Iterator
 from typing import Any
 
 import smart_open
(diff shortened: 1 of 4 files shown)


dsc/workflows/opencourseware.py
```diff
@@ -13,6 +14,234 @@
...
+                transformed_metadata[formatted_field_name] = field_method(source_metadata)
+            else:
+                transformed_metadata[formatted_field_name] = field_method()
+
+        return transformed_metadata
+
+    @classmethod
+    def dc_title(cls, source_metadata: dict) -> str:
+        """Build a title string from course numbers, title, and term year.
+
```
ehanson8, line 58: I can't remember what we had before for `dc.title` but the complexity of this definitely justifies a dedicated method!

dsc/workflows/opencourseware.py
```diff
@@ -13,6 +14,234 @@
...
+    @classmethod
+    def dc_coverage_temporal(cls, source_metadata: dict) -> str:
+        """Return a string combining term and year.
+
+        Example:
+            Input: {"term": "Spring", "year": "2024"}
+            Output: "Spring 2024"
+        """
+        return " ".join([source_metadata["term"], source_metadata["year"]])
```
ehanson8, line 216: Since `source_metadata["term"]` and `source_metadata["year"]` get joined a few times, you could consider whether it's worth doing it as an attribute

ehanson8 (approved):
This is great! One optional suggestion. Re: the potential tickets:

> Update codebase architecture to support metadata transformer class
There is appeal to abstracting this out but I also a bit hesitant given that many of these transformer methods are going to be `return source_metadata[field_name]` for most other workflows  So I would like to analyze how much lifting a metadata transformer class would do to see where its worth the trouble for a given workflow. 

>Define field method for dc.contributor.department in OCW transformer
As indicated in the mapping spreadsheet, the request is to use the provided department_numbers in the source metadata to look up the department string in DSpace. This suggests using the DSpace REST API to perform this lookup. We can continue our discussion in the ticket specific to this work (once it is created).

I think this should be DSS's responsibility and could be worked into the refactoring around DSpace 7, which will likely have a lot of this ID # resolution for configurable entities. I still think DSC should be isolated from DSpace and give DSS the ID #s to translate rather than reaching out itself.

dsc/workflows/opencourseware.py
```diff
@@ -13,6 +14,234 @@
...
+        "dc_language_iso",
+    ]
+
+    @classmethod
+    def transform(cls, source_metadata: dict) -> dict:
+        """Transform source metadata."""
+        transformed_metadata = {}
+        for field in cls.fields:
+            field_method = getattr(cls, field)
+            formatted_field_name = field.replace("_", ".")
```
ghukill, line 44: This derivation of a DSpace metadata field name, from a python method name, makes me a little nervous.  So far we're lucky that the metadata fields can be translated into a python method name, but it feels potentially brittle.

I had wondered if some kind of mapping could be used:

```python
from typing import Callable
    
    @classmethod
    def field_method_map(cls) -> list[tuple[Callable, str]]:
        return [
            (cls.dc_title, "dc.title"),
            (cls.dc_date_issued, "dc.date.issued"),
            (cls.dc_description_abstract, "dc.description.abstract"),
            (cls.dc_contributor_author, "dc.contributor.author"),
            (cls.creativework_learningresourcetype, "creativework.learningresourcetype"),
            (cls.dc_subject, "dc.subject"),
            (cls.dc_identifier_other, "dc.identifier.other"),
            (cls.dc_coverage_temporal, "dc.coverage.temporal"),
            (cls.dc_audience_educationlevel, "dc.audience.educationlevel"),
            (cls.dc_type, "dc.type"),
            (cls.dc_rights, "dc.rights"),
            (cls.dc_rights_uri, "dc.rights.uri"),
            (cls.dc_language_iso, "dc.language.iso"),
        ]
```

But I don't love the loss of a quick glance what fields we'll be targeting/returning.  And, it's a bit busy.

What if each field method returned a tuple, sometthing like `(<field_name>, <field_value>)`?

Also, I'm quite happy registering my nervousness and being okay with it!  We know this approach works for OCW, and if we hold off on registering a more generic `Transformer` class, then we could punt until then.

dsc/workflows/opencourseware.py
```diff
@@ -13,6 +14,234 @@
...
+
+            # check if 'source_metadata' is in signature
+            signature = inspect.signature(field_method)
+            if "source_metadata" in signature.parameters:
+                transformed_metadata[formatted_field_name] = field_method(source_metadata)
+            else:
+                transformed_metadata[formatted_field_name] = field_method()
```
ghukill, line 51: While this is _really_ cool, I wonder if it's necessary?  If we just pass `source_metadata` always, and methods can choose to ignore it, what's the harm?  Methods could have a leading underscore in the signature like this to avoid linting errors and indirectly suggest to developers its not used:

```python
def i_dont_need_metadata(_source_metadata:dict):
    ...
```

dsc/workflows/opencourseware.py
```diff
@@ -333,11 +333,13 @@ def item_metadata_iter(self) -> Iterator[dict[str, Any]]:
         ):
             yield {
                 "item_identifier": self.parse_item_identifier(file),
-                **self._extract_metadata_from_zip_file(file),
+                **self.metadata_transformer.transform(
+                    source_metadata=self._read_metadata_from_zip_file(file)
+                ),
```
ghukill, line 338: I had to sit with it, but I think this is a nice pattern of a) composing the _class_ not an _instance_ to `self`, then for each iteration here b) passing the source metadata to the class method `transform()`.  

Having them all be class methods probably makes them a _little_ trickier to work with (e.g. that do I want source metadata or not bit), but it makes it much simpler at this level.

ghukill (commented):
Overall, looks great to me.  Left a couple of questions, mostly around `OpenCourseWareTransformer.transform()`.  

In response to your PR questions:

> Update codebase architecture to support metadata transformer class

I think I agree with @ehanson8 here that it might be good to just sit with this for awhile.  If it _does_ turn out that OCW is the only one that requires this treatment, probably not worth abstracting out.  But if we enounter others, or it seems like other workflows might benefit from this pattern, then that abstraction work could just be rolled up into _that_ work.

> Define field method for dc.contributor.department in OCW transformer
As indicated in the mapping spreadsheet, the request is to use the provided department_numbers in the source metadata to look up the department string in DSpace. This suggests using the DSpace REST API to perform this lookup. We can continue our discussion in the ticket specific to this work (once it is created).

I wold love to talk about this more.

I think @ehanson8's proposal that DSC is fully air gapped from DSpace is a compelling one.  At the same time, I think there is a decent argument to be made that DSS gets increasingly less complex / important, to the point it's just a "ingest this item, as-is, to DSpace" application.  In which case, DSC would need to do this kind of work.  I must admit the latter appeals to me a bit, as it feels like metadata crafting work which I think of as DSC's bread and butter.

BUT, I'd hate for that interesting and complex discussion to take place here in a somewhat unrelated PR.  If a ticket was created to explore augmenting metadata creation here in DSC with data from DSpace (maybe keep it generic to "data" vs API calls)  then we could move that discussion there?

dsc/workflows/opencourseware.py
jonavellecuerdo: Yeah! I've been thinking of the `OpenCourseWareTransformer` as simply a collection of methods at its current state, since instantiation is not required. 🤔

dsc/workflows/opencourseware.py
jonavellecuerdo: I'm going to stick to it being 🆒 ! On a more serious note, I think the added code doesn't feel overwhelming or complex. 🤔 It's also the first time I've worked with the `inspect` module. We can revisit this `transform` method in the event we need to create another 'transformer' class.

jonavellecuerdo:
@ghukill @ehanson8 Thank you for your comments! A couple things worth noting:

* @ghukill I decided not to implement further changes.
* _Heard_ and holding off on a ticket to revisit codebase architecture. 👍🏼 
* Created a spike ticket to explore the use of DSpace data for transforms: [IN-1427](https://mitlibraries.atlassian.net/browse/IN-1427)

dsc/workflows/opencourseware.py
jonavellecuerdo: Keeping as-is for now!

PR #193, opened 2025-08-15
Add field method for 'dc.contributor.department' to OpenCourseWareTra…

jonavellecuerdo:
A mapping of department numbers to department name was provided, which was required to derive the `dc.contributor.department` field via the `OpenCourseWareTransformer`.

Review added [unit test](https://github.com/MITLibraries/dspace-submission-composer/pull/193/files#diff-66e8464388146924ee41c524d9617a44f16ffd3f5f334f2456cccc68c717d39cR98-R104).

NO

NO

- https://mitlibraries.atlassian.net/browse/IN-1428

### Developer

### Code Reviewer(s)

### DIFF
--- a/dsc/workflows/opencourseware.py
+++ b/dsc/workflows/opencourseware.py
@@ -3,7 +3,7 @@ import json
 import logging
 import zipfile
 from collections.abc import Iterable, Iterator
-from typing import Any
+from typing import Any, ClassVar
 
 import smart_open
 
@@ -23,6 +23,7 @@ class OpenCourseWareTransformer:
         "dc_date_issued",
         "dc_description_abstract",
         "dc_contributor_author",
+        "dc_contributor_department",
         "creativework_learningresourcetype",
         "dc_subject",
         "dc_identifier_other",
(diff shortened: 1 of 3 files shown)


dsc/workflows/opencourseware.py
```diff
@@ -35,6 +36,46 @@ class OpenCourseWareTransformer:
...
+        "CMS": "Massachusetts Institute of Technology. Program in Comparative Media Studies/Writing",  # noqa: E501
+        "HST": "Harvard University--MIT Division of Health Sciences and Technology",
+        "IDS": "Massachusetts Institute of Technology. Institute for Data, Systems, and Society",  # noqa: E501
+        "MAS": "Program in Media Arts and Sciences (Massachusetts Institute of Technology)",  # noqa: E501
+        "STS": "Massachusetts Institute of Technology. Program in Science, Technology and Society",  # noqa: E501
+        "ESD": "Massachusetts Institute of Technology. Engineering Systems Division",
+        "WGS": "MIT Program in Women's and Gender Studies",
+        "ESG": "MIT Experimental Study Group",
+        "EC": "Edgerton Center (Massachusetts Institute of Technology)",
+    }
```
ghukill, line 77: This has a bit of code smell to me... if nothing else those multiple `# noqa: E501`.

Ideally, I feel like this would be a JSON or CSV file.  It could be _read_ by this class, even still set as a class attribute.

This gets somewhat complicated, at least to me, because it would then kind of strongly suggest a workflow **directory** structure could be helpful, where it'd be natural to store this file in that workflow's directory.

Or.... at the very least I might suggest just skipping `E501` for this entire file with something like this at the top:

```
# ruff: noqa: E501
```

ghukill (commented):
Left a single comment about the mapping.

dsc/workflows/opencourseware.py
jonavellecuerdo: Hmm, I did ponder creating a file, but it opens the door to discussing the codebase architecture, which we decided to table for now. 🤔

Re: Skipping `E501` is okay as long as `black` runs beforehand to ensure `line-length` is adhered to 🤔. If the `pre-commit` hook is skipped (for whatever reason), developer needs to _remember_ to run `make black-apply`. User can ensure their IDE runs `black` when saving, but our setups may vary...

I like leaving rule `E501` on as it is a good nudge to do some cleanup, so I'm hesitant to ignore it at the module-level.

TLDR; Is it okay if we leave as-is? 😅

ghukill: You can also control `black` with a block like:

```python
# fmt: off
...
...
# fmt: on
```

We've done this in other projects for similar codeblocks.  

But yes!  **I'm good with as-is**.  Thanks for hearing me out, and sorry to hold things up.

ghukill (approved):

dsc/workflows/opencourseware.py
jonavellecuerdo: Interesting re: controlling `black`. I'll keep that in mind for the future!

PR #194, opened 2025-08-18
[DRAFT] Refactor 'reconcile' to operate at record-level

jonavellecuerdo:
_**WRITING IN PROGRESS**_

This PR restructures the `reconcile` methods implemented for the `SimpleCSV` and `OpenCourseWare` workflows to a record-oriented design. The key changed introduced in this PR was: **making `reconcile_items` an abstract method on the base `Workflow` class and removing `reconcile_metadata_and_bitstreams`**. Reasoning below:

   *  While there are _some_ similarities between the `SimpleCSV` and `OpenCourseWare` classes, there are enough differences that make it challenging to define a singular method that can encapsulate the required logic for each workflow. For instance:
      * [For `OpenCourseWare`, `reconcile` retrieves the item identifiers from the bitstream filenames](https://github.com/MITLibraries/dspace-submission-composer/pull/194/files#diff-5cea71db90920fb297d501bf66b83450316805c656241bbdd8a2f6e0c24e1228R327-R328). The naming convention for OCW zip files are consistent, making this possible.
         * [For OCW, "bitstreams without metadata" can be written to DynamoDB table.  ](https://github.com/MITLibraries/dspace-submission-composer/pull/194/files#diff-5cea71db90920fb297d501bf66b83450316805c656241bbdd8a2f6e0c24e1228R355-R360)
      * [For `SimpleCSV`, `reconcile` retrieves the item identifiers from a provided metadata CSV file](https://github.com/MITLibraries/dspace-submission-composer/pull/194/files#diff-e1df9d0c01a124123cf096119c5185ec6a442b46810342270af46199d254cbf4R49). Since there is no set naming convention for bitstream filenames,  it may be difficult to parse what is the "item identifier" (item identifiers can vary in length) vs "sequence number" (if it even has one)
         * For the reason above, "bitstreams without metadata" are not written to DynamoDB table.

NO

NO

- https://mitlibraries.atlassian.net/browse/IN-1408

### Developer

### Code Reviewer(s)

### DIFF
--- a/dsc/workflows/base/__init__.py
+++ b/dsc/workflows/base/__init__.py
@@ -138,7 +138,7 @@ class Workflow(ABC):
         self.exclude_prefixes: list[str] = ["archived/", "dspace_metadata/"]
 
         # cache list of bitstream uris
-        self.batch_bitstreams = self.get_batch_bitstreams()
+        self._batch_bitstreams = None  # self.get_batch_bitstreams()
 
         # capture high-level results of executed commands
         self.submission_summary: dict[str, int] = {
@@ -216,7 +216,9 @@ class Workflow(ABC):
         """
 
     def reconcile_items(self):
-        reconciled = {}  # dict where keys = item_identifier, values = bitstream uris
+        reconciled_items = (
+            {}
+        )  # dict where keys = item_identifier, values = bitstream uris
         bitstreams_without_metadata = []
         metadata_without_bitstreams = []
 
@@ -263,8 +265,9 @@ class Workflow(ABC):
             )
 
         # check for unmatched bitstreams
+        matched_bitstreams = reconciled.values()
         bitstreams_without_metadata.extend(
-            list(set(self.batch_bitstreams) - set(itertools.chain(*reconciled.values())))
+            list(set(self.batch_bitstreams) - set(itertools.chain(*matched_bitstreams)))
         )
 
         reconcile_summary = {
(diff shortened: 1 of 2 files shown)


ghukill (commented):
Thanks for putting together a PR to discuss.  

After taking a first read / pass, I feel like this refactor is taking steps in the right direction but maybe could still go further?

As a hypothetical, a few questions:

1. Would both these workflows benefit from a **singular** `reconcile_item()` method?  and then either the workflows have the **plural, orchestrator**  method of `reconcile_items()`... or maybe that moves back to the base class?

2. Why does `SimpleCSV` loop through metadata items, but `OCW` loops  through bitstreams?  Is it because bitstreams _are_ intellectual items in `OCW`?  Could that be normalized more, such that both could share a loop/orchestrator signature?   and again, consider moving that back to the base workflow?

--- 

#### Some Misc Comments / Observations

1- In `simple_csv.py`, roughly lines 102-143, though I haven't gone through with the finest of combs it feels like a _lot_ of code that kind of keep doing the same thing.  

I would suggest we either do away with the workflow events _entirely_, or lean into them _heavily_.  If we think of them as "bookeeping", then perhaps we could/should write to them directly and earlier in the method?  Then just analyze those workflow events for reporting and logging?

That might support doing away with the assymetrically named `reconciled_items` and `metadata_without_bitstreams`.  

--- 

Starting with these questions / comments for now.

jonavellecuerdo:
Closing this PR and will open a new PR with logically ordered commits.

PR #195, opened 2025-08-20
Align workflow methods for iterating over item metadata

jonavellecuerdo:

This PR is part of a series of updates related to aligning the structure of the `reconcile` method with the other DSC CLI commands (`submit` and `finalize`). Specifically, this PR aligns the method signatures of the `item_metadata_iter` methods. For the OpenCourseWare (OCW) workflow, the method would previously raise a `FileNotFoundError` if a metadata JSON file (`data.json`) was not found in the zip file. These changes ensure that OCW returns a dict regardless (containing at least the item identifier). 

This **also fixes a breaking change** that was introduced by the previous update that introduced a metadata transformer for OCW (https://github.com/MITLibraries/dspace-submission-composer/pull/192). During that implementation, I missed that the metadata mapping JSON file was still required as all workflows create DSpace metadata (i.e., "transformed source metadata") through the shared `ItemSubmission.create_dspace_metadata` method. I opted to update the metadata mapping JSON file for now to narrow the scope of this PR.

**NOTE:** In my discussions with @ghukill, we originally thought we could take a step further with `OpenCourseWare.item_metadata_iter` and ensure the method returns the "source metadata" (pre-transformation), however, due to the same reason above, this would require changing how/when we create DSpace metadata. I strongly suggest tabling this discussion for later and focusing on **aligning the structure of the `reconcile` command** for now.

Review added unit test.

NO

NO

* https://mitlibraries.atlassian.net/browse/IN-1431

### DIFF
--- a/dsc/workflows/opencourseware.py
+++ b/dsc/workflows/opencourseware.py
@@ -342,7 +342,7 @@ class OpenCourseWare(Workflow):
         for file in s3_client.files_iter(
             bucket=self.s3_bucket, prefix=self.batch_path, file_type=".zip"
         ):
-            item_identifier = self.parse_item_identifier(file)
+            item_identifier = self._parse_item_identifier(file)
 
             try:
                 self._read_metadata_from_zip_file(file)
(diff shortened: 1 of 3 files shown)


dsc/workflows/opencourseware.py
```diff
@@ -381,20 +381,31 @@ def reconcile_bitstreams_and_metadata(self) -> bool:
...
         """
         s3_client = S3Client()
         for file in s3_client.files_iter(
             bucket=self.s3_bucket, prefix=self.batch_path, file_type=".zip"
         ):
+            try:
+                source_metadata = self._read_metadata_from_zip_file(file)
+                transformed_metadata = self.metadata_transformer.transform(
+                    source_metadata
+                )
```
ghukill, line 402: Could this transformation move outside of this `try/except` block, which looks to only capture `FileNotFoundError`?

ghukill (approved):
Left a question about moving the `self.metadata_transformer.transform(source_metadata)` outside of the `try/except` block, but my understanding is that this PR is a building block towards additional work.

By reintroducing the field mapping support for OCW, I'm assuming this sidesteps some errors and allows focusing on continuing with the refactor work.

ehanson8 (approved):
A good set of changes for better alignment between the workflow classes!

dsc/workflows/opencourseware.py
jonavellecuerdo: Good catch! Updating. :)

PR #198, opened 2025-08-25
Add ArchivesSpace workflow

ehanson8:
Add `ArchivesSpace` workflow and associated functionality.

I originally was thinking I needed a more invasive approach for generating the ingest report but realize this could all be accomplished with the DynamoDB data. To ensure DSC is only generating a report for items that were successfully ingested on the current run, I’m using the following criteria: 

- The ingest report runs after all results messages are processed and all item submissions are updated in DynamoDB 
- `status` in DynamoDB is set to `INGEST_SUCCESS `
- `run_date` in DynamoDB matches `self.run_date` to ensure it was updated in the current run 

The  check on line 500 of `dsc/workflows/base/__init__.py` ensures anything already set to `INGEST_SUCCESS` will be skipped before the `last_run_date` is set on line 528. Please let me know if there holes in this logic!

Not possible at this stage

NO

NO

* https://mitlibraries.atlassian.net/browse/IN-1100

### Developer

### Code Reviewer(s)

### DIFF
--- a/dsc/item_submission.py
+++ b/dsc/item_submission.py
@@ -49,6 +49,7 @@ class ItemSubmission:
     batch_id: str
     item_identifier: str
     workflow_name: str
+    source_system_identifier: str | None = None
     collection_handle: str | None = None
     last_run_date: datetime | None = None
     submit_attempts: int = 0
@@ -67,7 +68,11 @@ class ItemSubmission:
 
     @classmethod
     def get_or_create(
-        cls, batch_id: str, item_identifier: str, workflow_name: str
+        cls,
+        batch_id: str,
+        item_identifier: str,
+        workflow_name: str,
+        source_system_identifier: str | None = None,
     ) -> ItemSubmission:
         """Get or create an ItemSubmission.
 
@@ -75,7 +80,10 @@ class ItemSubmission:
         finds a corresponding record.
         """
         return cls.get(batch_id, item_identifier) or cls.create(
-            batch_id, item_identifier, workflow_name
+            batch_id=batch_id,
+            item_identifier=item_identifier,
+            workflow_name=workflow_name,
+            source_system_identifier=source_system_identifier,
         )
 
     @classmethod
@@ -107,10 +115,19 @@ class ItemSubmission:
 
     @classmethod
     def create(
-        cls, batch_id: str, item_identifier: str, workflow_name: str
+        cls,
+        batch_id: str,
+        item_identifier: str,
+        workflow_name: str,
+        source_system_identifier: str | None = None,
     ) -> ItemSubmission:
         """Create an ItemSubmission."""
-        return cls(batch_id, item_identifier, workflow_name)
+        return cls(
+            batch_id=batch_id,
+            item_identifier=item_identifier,
+            workflow_name=workflow_name,
+            source_system_identifier=source_system_identifier,
+        )
 
     @classmethod
     def _from_db(cls, item_submission_db: ItemSubmissionDB) -> ItemSubmission:
--- a/dsc/workflows/archivesspace.py
+++ b/dsc/workflows/archivesspace.py
@@ -1,6 +1,6 @@
-import csv
 import logging
 
+import pandas as pd
 import smart_open
 
 from dsc.db.models import ItemSubmissionStatus
@@ -41,12 +41,13 @@ class ArchivesSpace(SimpleCSV):
         handle_uri_mapping = {}
 
         # find item submissions that were successfully ingested on the current run
-        for item_submission in [
+        successful_item_submissions: list[ItemSubmission] = [
             item_submission
             for item_submission in ItemSubmission.get_batch(self.batch_id)
             if item_submission.last_run_date == self.run_date
             and item_submission.status == ItemSubmissionStatus.INGEST_SUCCESS
-        ]:
+        ]
+        for item_submission in successful_item_submissions:
             handle_uri_mapping[item_submission.source_system_identifier] = (
                 item_submission.dspace_handle
                 if item_submission.dspace_handle
@@ -58,14 +59,26 @@ class ArchivesSpace(SimpleCSV):
             )
             return
 
+        # create dataframe
+        df = pd.DataFrame(
+            [
+                {
+                    "ao_uri": item.source_system_identifier,
+                    "dspace_handle": (
+                        item.dspace_handle
+                        if item.dspace_handle
+                        else "DSpace handle not set, possible error"
+                    ),
+                }
+                for item in successful_item_submissions
+            ]
+        )
+
         # create report in S3 bucket
         with smart_open.open(
             f"s3://{self.output_path}/{self.batch_id}-{run_date_str}.csv", "w"
         ) as csv_file:
-            writer = csv.writer(csv_file)
-            writer.writerow(["ao_uri", "dspace_handle"])
-            for ao_uri, dspace_handle in handle_uri_mapping.items():
-                writer.writerow([ao_uri, dspace_handle])
+            df.to_csv(csv_file, index=False)
 
         logger.debug(
             f"Completed ingest report for batch '{self.batch_id}' on run date"
--- a/dsc/workflows/archivesspace.py
+++ b/dsc/workflows/archivesspace.py
@@ -1,0 +1,73 @@
+import csv
+import logging
+
+import smart_open
+
+from dsc.db.models import ItemSubmissionStatus
+from dsc.item_submission import ItemSubmission
+from dsc.workflows.base.simple_csv import SimpleCSV
+
+logger = logging.getLogger(__name__)
+
+
+class ArchivesSpace(SimpleCSV):
+    """Workflow for ArchivesSpace deposits.
+
+    The deposits managed by this workflow are requested by the Department of Distinctive
+    Collections (DDC) and are submitted to Dome. This workflow utilizes the
+    workflow_specific_processing() method to produce a report for linking ArchivesSpace
+    records to the newly-ingested Dome items. The report is sent to an S3 bucket in the
+    output_path property after it is generated.
+    """
+
+    workflow_name: str = "archivesspace"
+
+    @property
+    def metadata_mapping_path(self) -> str:
+        return "dsc/workflows/metadata_mapping/archivesspace.json"
+
+    @property
+    def output_path(self) -> str:
+        return "output-bucket"
+
+    def workflow_specific_processing(self) -> None:
+        """Generate an ingest report linking DSpace handles to ArchivesSpace URIs.
+
+        This report is used in a separate process for updating metadata records in
+        ArchivesSpace with the newly-created DSpace handles.
+        """
+        run_date_str = self.run_date.strftime("%Y-%m-%d-%H:%M:%S")
+
+        handle_uri_mapping = {}
+
+        # find item submissions that were successfully ingested on the current run
+        for item_submission in [
+            item_submission
+            for item_submission in ItemSubmission.get_batch(self.batch_id)
+            if item_submission.last_run_date == self.run_date
+            and item_submission.status == ItemSubmissionStatus.INGEST_SUCCESS
+        ]:
+            handle_uri_mapping[item_submission.source_system_identifier] = (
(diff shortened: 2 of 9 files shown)


dsc/workflows/archivesspace.py
```diff
@@ -0,0 +1,73 @@
...
+    output_path property after it is generated.
+    """
+
+    workflow_name: str = "archivesspace"
+
+    @property
+    def metadata_mapping_path(self) -> str:
+        return "dsc/workflows/metadata_mapping/archivesspace.json"
+
+    @property
```
ehanson8, line 29: This will undoubtedly be updated once the infra is figured out so just consider this a placeholder

tests/conftest.py
```diff
@@ -123,6 +124,12 @@ def simple_csv_workflow_instance(metadata_mapping):
     return TestSimpleCSV(batch_id="batch-aaa")
 
 
+@pytest.fixture
+@freeze_time("2025-01-01 09:00:00")
```
ehanson8, line 128: Needed for the unit tests given the reliance on `self.run_date`

dsc/workflows/archivesspace.py
```diff
@@ -0,0 +1,73 @@
...
+        for item_submission in [
+            item_submission
+            for item_submission in ItemSubmission.get_batch(self.batch_id)
+            if item_submission.last_run_date == self.run_date
+            and item_submission.status == ItemSubmissionStatus.INGEST_SUCCESS
+        ]:
+            handle_uri_mapping[item_submission.source_system_identifier] = (
+                item_submission.dspace_handle
+                if item_submission.dspace_handle
+                else "DSpace handle not set, possible error"
```
ehanson8, line 53: I’m not including the possibility that items were sent to a DSpace submission queue (which does not create a handle) because DDC has never used them

dsc/workflows/archivesspace.py
```diff
@@ -0,0 +1,73 @@
...
+
+
+class ArchivesSpace(SimpleCSV):
+    """Workflow for ArchivesSpace deposits.
+
+    The deposits managed by this workflow are requested by the Department of Distinctive
+    Collections (DDC) and are submitted to Dome. This workflow utilizes the
+    workflow_specific_processing() method to produce a report for linking ArchivesSpace
+    records to the newly-ingested Dome items. The report is sent to an S3 bucket in the
+    output_path property after it is generated.
```
ghukill, line 20: Nice docstring!

dsc/workflows/archivesspace.py
ghukill: And just to confirm, this would need to be dynamic, e.g. an env var, right?

dsc/workflows/archivesspace.py
```diff
@@ -0,0 +1,73 @@
...
+
+        handle_uri_mapping = {}
+
+        # find item submissions that were successfully ingested on the current run
+        for item_submission in [
+            item_submission
+            for item_submission in ItemSubmission.get_batch(self.batch_id)
+            if item_submission.last_run_date == self.run_date
+            and item_submission.status == ItemSubmissionStatus.INGEST_SUCCESS
+        ]:
```
ghukill, line 49: I might propose that this somewhat complex list comprehension is assigned to a variable first.  In doing so, we can get two things: 1) a type of the list, e.g. `list[ItemSubmission]` and 2) the actual name of that variable is semantically meaningful.

dsc/workflows/archivesspace.py
```diff
@@ -0,0 +1,73 @@
...
+        """
+        run_date_str = self.run_date.strftime("%Y-%m-%d-%H:%M:%S")
+
+        handle_uri_mapping = {}
+
+        # find item submissions that were successfully ingested on the current run
+        for item_submission in [
+            item_submission
+            for item_submission in ItemSubmission.get_batch(self.batch_id)
+            if item_submission.last_run_date == self.run_date
```
ghukill, line 47: If `last_run_date` is second level granularity.... then I dig it!  But if it's only `YYYY-MM-DD`, we might have conflicts?

Maybe related, what sets `ItemSubmission.last_run_date` in the DB?  Does reconcile, submit, and finalize all set it?

Now that I read this more closely, does this work because:  `self.run_date` was minted by _this_ workflow instance for _this_ run which just wrote those items to the DB, and so it should match precisely?

dsc/workflows/archivesspace.py
```diff
@@ -0,0 +1,73 @@
...
+            return
+
+        # create report in S3 bucket
+        with smart_open.open(
+            f"s3://{self.output_path}/{self.batch_id}-{run_date_str}.csv", "w"
+        ) as csv_file:
+            writer = csv.writer(csv_file)
+            writer.writerow(["ao_uri", "dspace_handle"])
+            for ao_uri, dspace_handle in handle_uri_mapping.items():
+                writer.writerow([ao_uri, dspace_handle])
```
ghukill, line 68: To each their own, but pandas could be an easier way to generate a CSV?

One thing that appeals to me about that approach is the effort is directed at crafting a dataframe that has your final desired structure.  The generation + writing of the CSV is kind of a minor.  For debugging purposes, it's easy to even print that dataframe to see what's going on.  Said another way, it exists before write which can be helpful for debugging.

But totally optional.

dsc/item_submission.py
```diff
@@ -49,6 +49,7 @@ class ItemSubmission:
     batch_id: str
     item_identifier: str
     workflow_name: str
+    source_system_identifier: str | None = None
```
ghukill, line 52: Picking this line  somewhat arbitrarily, but wondering if you could talk a bit more about the `source_system_identifier`?

I noticed this trickles all the way into DynamoDB.  

FWIW, I'm a _big_ fan of storing as many identifiers as possible, but wondering what about this workflow made this seem important now versus other workflows.  

I do like the name though: think it's very clear, very direct.

ghukill (commented):
I'm unsure if I'm formerly requesting any changes, but had a few comments and questions.

For the most part, I'm following along and looks pretty good!  I think how small the `ArchivesSpace` workflow class is speaks to some of overall pattern of workflows extending "base" ones.

dsc/workflows/archivesspace.py
ehanson8: Yes, almost certainly an env var

dsc/workflows/archivesspace.py
ehanson8: Good call, will do!

dsc/item_submission.py
ehanson8: If we need to provide a link from the source system (here an archival object URI in ArchivesSpace) to the DSpace handle, we would use `source_system_identifier`. The `ArchivesSpace` workflow is the first with a defined use case for it. With the previous DDC deposits, I use this report to run a batch process that creates a new digital object with the DSpace handle attached to specified archival object. 

I kept the name generic since the Imaging Lab or other future stakeholders might have a source system where they need to track DSpace handles for content that was deposited.

dsc/workflows/archivesspace.py
ehanson8: I need to take advantage of `pandas` more often, I'll update!

dsc/item_submission.py
ehanson8: That was the reason for the default of `None`, only need to supply when it'll actually be used as it is for `ArchivesSpace`

jonavellecuerdo: In ArchivesSpace, is `item_identifier` conceptually different from a URI (e.g., is it a separate column in the metadata)? 🤔 What I'm trying to get at is whether the "source system identifier" could be used as the `item_identifier` for DSC?

Are there any drawbacks/complications if we made that change? If not, this would remove the need to introduce a new column. 🤔

ehanson8: Good question! That might work in some other workflows but I don't think for this one, the `source_system_identifier` would be `/repositories/2/archival_objects/458300` (needed for updating ASpace) and the `item_identifier` would be `02-000458300` (naming convention coming out of the digitization process used for the bitstreams). I can easily see a future workflow where both identifiers are aligned in a different system, but not this one.

ehanson8:
Pushed a commit with the discussed updates

dsc/item_submission.py
jonavellecuerdo: Hmm, does the string `/repositories/2/archival_objects/458300` always link back to the same item as `02-000458300`? 🤔 

Essentially, from DSC's perspective, it only needs a unique identifier for the item both those identifiers are referring to. Whether it's `02-000458300` or `/repositories/2/archival_objects/458300`, it just needs one "item identifier".

ghukill: I guess I'd have to defer entirely to @ehanson8 here!  Feels like a wrinkle of the ASpace workflow.

Conceptually, I can appreciate, and think we should support, when there is a "source system" identifier that does not match the "DSC / working item identifier" that stakeholders would like to provide.

I'll admit the latter, the "DSC / working item identifier" has always been a little odd, given that it really is kind of _only_ for the DSC/DSS space.  

But I could see a universe where the "source" identifier is something odd / long / weird, has unique characters, who knows.  But the batch we are given in DSC, maybe it's just an incrementing number or something.  

I would _also_ not see any issue if they sometimes were the same.  Or, as @ehanson8 points out, that often workflows won't have a "source" identifier.

ghukill: I mean, maybe we should think of the DSC `item_identifier` more like `batch_item_identifier` which makes it clear it's an identifier that needs to be unique in the batch, for a workflow, but that's kind of the end of it's responsibility.  

Noting that `item_identifier` can be repeated, right?  like in a secondary, later batch?  that would effectively update the item in DSpace?

ehanson8: > `item_identifier` would be `02-000458300` (naming convention coming out of the digitization process used for the bitstreams).

This is the important part, DSC can't find the bitstreams if we are not using this identifier. The `source_system_identifier` (the archival object URI) serves no purpose to DSC but is needed for updating ArchivesSpace with the ingest report

dsc/workflows/archivesspace.py
```diff
@@ -0,0 +1,86 @@
...
+
+        This report is used in a separate process for updating metadata records in
+        ArchivesSpace with the newly-created DSpace handles.
+        """
+        run_date_str = self.run_date.strftime("%Y-%m-%d-%H:%M:%S")
+
+        handle_uri_mapping = {}
+
+        # find item submissions that were successfully ingested on the current run
+        successful_item_submissions: list[ItemSubmission] = [
```
ghukill, line 44: Thanks for list comprehension --> variable!  Helps my reading of the code.

dsc/workflows/archivesspace.py
```diff
@@ -58,14 +59,26 @@ def workflow_specific_processing(self) -> None:
...
+
         # create report in S3 bucket
         with smart_open.open(
             f"s3://{self.output_path}/{self.batch_id}-{run_date_str}.csv", "w"
         ) as csv_file:
-            writer = csv.writer(csv_file)
-            writer.writerow(["ao_uri", "dspace_handle"])
-            for ao_uri, dspace_handle in handle_uri_mapping.items():
-                writer.writerow([ao_uri, dspace_handle])
+            df.to_csv(csv_file, index=False)
```
ghukill, line 81: Awesome!  I forget that we can use `smart_open` to get a file-like object and then pass that to `df.to_csv()`.  Very nice!

ghukill (approved):

dsc/item_submission.py
ehanson8: @jonavellecuerdo is correct ⬆️

dsc/workflows/metadata_mapping/archivesspace.json
```diff
@@ -0,0 +1,21 @@
+{
+  "dc.title": {
+    "source_field_name": "title",
+    "language": "en_US"
```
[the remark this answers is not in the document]
ehanson8: Actually, I will just remove it from `dc.title` and leave it on `rights_statement` and `description` since those are boilerplate strings from DDC, if that's OK @jonavellecuerdo

PR #200, opened 2025-09-16
Add batch creation method to Workflow abstract base class 

jonavellecuerdo:
Support DSC workflows that require programmatic batch creation.

For now, passing unit tests will suffice.

NO

YES - Enables the addition of a "batch creation" step to the DSC workflow.

- https://mitlibraries.atlassian.net/browse/IN-1455

### DIFF
--- a/dsc/workflows/base/__init__.py
+++ b/dsc/workflows/base/__init__.py
@@ -14,6 +14,7 @@ import jsonschema.exceptions
 from dsc.config import Config
 from dsc.db.models import ItemSubmissionDB, ItemSubmissionStatus
 from dsc.exceptions import (
+    BatchCreationFailedError,
     InvalidSQSMessageError,
     InvalidWorkflowNameError,
     ReconcileFailedError,
@@ -222,6 +223,60 @@ class Workflow(ABC):
         """
 
     @final
+    def create_batch(self) -> None:
+        """Create a batch of item submissions for processing.
+
+        A "batch" refers to a collection of item submissions that are grouped together
+        for coordinated processing, storage, and workflow execution. Each batch
+        typically consists of multiple items, each with its own metadata and
+        associated files, organized under a unique batch identifier.
+
+        This method prepares the necessary assets in S3 (programmatically as needed)
+        and records each item in the batch to DynamoDB.
+        """
+        item_submissions, errors = self.prepare_batch()
+        if errors:
+            raise BatchCreationFailedError
+        self._create_batch_in_db(item_submissions)
+
+    @abstractmethod
+    def prepare_batch(self) -> tuple[list, ...]:
+        """Prepare batch submission assets in S3.
+
+        This method performs the required steps to prepare a batch
+        of item submissions in S3. These steps must include (at minimum)
+        the following checks:
+
+        - Check if there is metadata for the item submission;
+          otherwise raise dsc.exceptions.ItemMetadataNotFoundError
+        - Check if there are any bitstreams for the item submission;
+          otherwise raise dsc.exceptions.ItemBitstreamsNotFoundError
+
+        MUST be overridden by workflow subclasses.
+
+        Returns:
+            A tuple of item submissions (init params) represented as a
+            list of dicts and errors represented as a list of tuples
+            containing the item identifier and the error message.
+        """
+        pass  # noqa: PIE790
+
+    @final
+    def _create_batch_in_db(self, item_submissions: list[dict]) -> None:
+        """Write records for a batch of item submissions to DynamoDB.
+
+        This method loops through the item submissions (init params)
+        represented as a list dicts. For each item submission, the
+        method creates an instance of ItemSubmission and saves the
+        record to DynamoDB.
+        """
+        for item_submission_init_params in item_submissions:
+            item_submission = ItemSubmission.create(**item_submission_init_params)
+            item_submission.last_run_date = self.run_date
+            item_submission.status = ItemSubmissionStatus.BATCH_CREATED
+            item_submission.save()
+
+    @final
     def reconcile_items(self) -> bool:
         """Reconcile item submissions for a batch.
 
@@ -241,6 +296,8 @@ class Workflow(ABC):
         NOTE: This method is likely the first time a record will be inserted
         into DynamoDB for each item submission. If already present,
         its status will be updated.
+
+        TODO: Reconcile methods will be deprecated after end-to-end testing.
         """
         reconciled_items = {}  # key=item_identifier, value=list of bitstream URIs
         bitstreams_without_metadata = []  # list of bitstream URIs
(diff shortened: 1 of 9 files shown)


dsc/workflows/base/__init__.py
```diff
@@ -248,13 +290,15 @@ def reconcile_items(self) -> bool:
 
         # loop through each item metadata
         for item_metadata in self.item_metadata_iter():
-            item_submission = ItemSubmission.get_or_create(
+            item_submission = ItemSubmission.get(
```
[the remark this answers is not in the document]
jonavellecuerdo: Resolving this comment and hoping we can continue the discussion after review of additional commits. See below!

jonavellecuerdo:
[DRAFT]

The latest commits were motivated by the discussion in https://github.com/MITLibraries/dspace-submission-composer/pull/200#discussion_r2356510074. In order to support moving the `item_metadata_iter` into the batch creation step--**specifically, during the writing of records to DynamoDB**, the following updates were implemented:
* [Storing the source metadata as a serialized JSON in the DynamoDB table](https://github.com/MITLibraries/dspace-submission-composer/pull/200/files#diff-ac3b91a3d46f6b525e09196ae93a47ac7d830e62ae08494b8d8d90e47f46bc98R105)
   *  [Adding a `source_metadata` property to `ItemSubmission` that parses the serialized JSON as needed](https://github.com/MITLibraries/dspace-submission-composer/pull/200/files#diff-99c0807f75c8ef71eef8d7a50e3518ff669cfde4a65aa443245cf99658e15647R75-R80)
* [Setting the `bitstream_s3_uris` attribute during the batch creation step](https://github.com/MITLibraries/dspace-submission-composer/pull/200/files#diff-9409cf7165577c3250e707080466409d47908aae682829ddb8a1abadecda795aR264-R266)
* Updating the methods for creating DSpace metadata in `ItemSubmission` to access source metadata from `self`

ehanson8:
> If the goal is just to avoid S3 reads, I'm not sure Dynamo is a good fit. If _part_ of the goal is to normalize how different workflows loop through metadata, that feels a bit different. Happy to chat it out if helpful, sync or async, but I do worry that Dynamo isn't up for the task of storing item metadata.

I was surprised by this as well. I know I was saying to "minimize calls to `item_metadata_iter()`" but I was more thinking about driving `reconcile` with DynamoDB post-`create_batch` rather than moving source metadata and bitstreams functionality away from `submit`. I don't know that `create_batch` needs to be handling that, so I think this can be a simpler set of changes just related to `reconcile`

ehanson8:
> My answer would be **ignoring loose files**. 

I agree, this is a minimal risk in the first place but we could have a ticket to address the issue later

jonavellecuerdo:
@ghukill @ehanson8 I am requesting for another pass at review! As mentioned, I dropped the two commits that involved changes to the DynamoDB model. I've refactored the batch creation methods of `Workflow.prepare_batch` and `Workflow._create_batch_in_db`. While we discussed holding off on touching subclasses, I did find it helpful to go ahead and derive `SimpleCSV.prepare_batch` (it really wasn't a lot of code)!

Looking forward to getting your thoughts. :)

dsc/workflows/base/__init__.py
```diff
@@ -221,6 +222,58 @@ def item_metadata_iter(self) -> Iterator[dict[str, Any]]:
...
+        """Create a batch of item submissions for processing.
+
+        A "batch" refers to a collection of item submissions that are grouped together
+        for coordinated processing, storage, and workflow execution. Each batch
+        typically consists of multiple items, each with its own metadata and
+        associated files, organized under a unique batch identifier.
+
+        This method prepares the necessary assets in S3 (programmatically as needed)
+        and records each item in the batch to DynamoDB.
+        """
```
ghukill, line 235: Just observing this is a nice, informative docstrsing!

dsc/workflows/base/__init__.py
```diff
@@ -221,6 +222,58 @@ def item_metadata_iter(self) -> Iterator[dict[str, Any]]:
...
+        item_submissions, errors = self.prepare_batch()
+        if errors:
+            raise BatchCreationFailedError
```
ghukill, line 238: Is this a bit of a stubbed exception handling?  FWIW, totally fine with that.  I ask, because it feels like we're not doing anything with the `errors` output which could be helpful for reporting.

dsc/workflows/base/simple_csv.py
```diff
@@ -92,3 +95,47 @@ def item_metadata_iter(
...
+        return item_submission_init_params, errors
+
+    def _insufficient_item_metadata(self, item_metadata: dict) -> bool:
+        """Check if the metadata for the item is sufficient.
+
+        All item metadata will have a set value for the 'item_identifier' field.
+        If the count of non-null fields is exactly 1, this means only the item
+        identifier has been provided; therefore the metadata is insufficient.
+        """
+        return sum(1 for value in item_metadata.values() if value is not None) == 1
```
ghukill, line 141: I can't pinpoint why, but this makes me a _litte_ uneasy?  But doesn't feel worth blocking the PR.  I _do_ like there is a place for this kind of validation, and can always be revisited later if need  be.  I like that this suggests a pattern other workflows can follow.

dsc/workflows/base/__init__.py
jonavellecuerdo: Docstring was generated with the help of Copilot extension on VSCode. 🤓

tests/conftest.py
```diff
@@ -24,6 +24,7 @@
 from dsc.workflows.base.simple_csv import SimpleCSV
 
 
+# Test Workflow classes ######################
```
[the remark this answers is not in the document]
jonavellecuerdo: Inspired from the `transmogrifier/conftest.py` -- Yay us!

dsc/workflows/base/simple_csv.py
jonavellecuerdo: Tagging @ehanson8 here as it relates to (https://github.com/MITLibraries/dspace-submission-composer/pull/200#discussion_r2376926643) as well. 

For clarification, this is the check to see if there is metadata for the item submission. 
The idea was: 
* The `item_metadata_iter` methods for `SimpleCSV` and `OpenCourseWare` can return a dictionary containing only `item_identifier`, but as shown in `SimpleCSV.prepare_batch` the metadata must comprise of _more than just the `item_identifier`_. 

Given this information, what do ya'll think? Do you have suggestions for a function rename?

ehanson8: I'm just not sure it's really necessary unless we're going to do it more robustly. DSpace is going to kick out anything without a title, so that's effectively guaranteed later in the process. Also I think the risk of having an identifier but absolutely nothing else is minimal (except in the cases where we have to go grab a metadata file)

ehanson8: Also, I'm not sure this is where we would even want to do this check. We read the source metadata when we create the DSpace metadata so that might be a better place for any eventual validation

jonavellecuerdo: @ghukill To clarify, in the case of Wiley this check would still be in `prepare_batch`?

ehanson8 (approved):
Excellent work!

PR #201, opened 2025-10-01
In 1460 add batch creation report

jonavellecuerdo:
Reporting is required for the recently added batch creation step, which assumes the responsibility of recording
item submissions to DynamoDB (previously handled by the reconcile step). This also marks an important shift to retrieve reporting data from DynamoDB instead of the WorkflowEvents module.

As work to migrate DSC -> DSO (infrastructure) begins, it might be tricky to test sending of actual emails. Review of new unit tests is sufficient.

NO - The first commit ignores the vulnerability raised by `pip-audit`.

NO

- https://mitlibraries.atlassian.net/browse/IN-1460

### DIFF
--- a/dsc/reports/create_batch.py
+++ b/dsc/reports/create_batch.py
@@ -1,0 +1,117 @@
+from __future__ import annotations
+
+import datetime
+import logging
+from io import StringIO
+from typing import TYPE_CHECKING, ClassVar
+
+import boto3
+import pandas as pd
+from jinja2 import Environment, FileSystemLoader, Template, select_autoescape
+
+from dsc.config import Config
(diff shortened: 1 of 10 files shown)


dsc/reports/create_batch.py
```diff
@@ -0,0 +1,117 @@
...
+
+    fields: ClassVar[list[str]] = [
+        "batch_id",
+        "item_identifier",
+        "source_system_identifier",
+        "status",
+        "status_details",
+        "dspace_handle",
+        "ingest_date",
+    ]
```
jonavellecuerdo, line 43: I don't expect _all_ attributes are needed for the CSV files created by the report classes.

PR #225, opened 2026-04-03
Prerequisites for Digitized Theses workflow

jonavellecuerdo:

This PR comprises of a series of updates to the DSC app in preparation for the Digitized Theses workflow. At high level, these updates are:

* [Updating `Workflow.prepare_batch` to return `ItemSubmission` objects](https://github.com/MITLibraries/dspace-submission-composer/pull/225/commits/d9911d58de02355da637c4abe24ed481a15603b9)
   * This change was proposed in initial DSC refactoring discussions
   * Prior to this change, this method was returning _attributes_ for an `ItemSubmission` object
* [Create a util function for invoking `sync` AWS CLI command](https://github.com/MITLibraries/dspace-submission-composer/pull/225/commits/f62082da8b148f173c6841117539a3104e06ad0f)
   * Called by `sync` DSC CLI command
   * To be called by digitized-theses workflow during batch creation when syncing batch folder from `digitized-theses-workspace` to a minted batch folder in `dsc-upload` in S3.
* [Rework `archive_file_with_new_key` as general method for moving files in S3](https://github.com/MITLibraries/dspace-submission-composer/pull/225/commits/17c94a11c37e1d7c03795c53aff3ac0e33a6d824)
   * This method is not currently used by any implemented workflows (pulled from Wiley)
   * This method will be used by digitized-theses workflow to move files into `replacement-theses/` and `new_theses/` subfolders (prefixes)
* [Make collection handle an optional argument for `submit` CLI command](https://github.com/MITLibraries/dspace-submission-composer/pull/225/commits/ddbe504a0918851258c3867b744b0f04c7cb9018)

Full DSC `OpenCourseWare` workflow run:
   1. create ([step function execution](https://us-east-1.console.aws.amazon.com/states/home?region=us-east-1#/v2/executions/details/arn:aws:states:us-east-1:222053980223:execution:dso-dev:d708737a-7c9d-42ea-a74e-9d4b6a6fcb1a), [CloudWatch log stream](https://us-east-1.console.aws.amazon.com/cloudwatch/home?region=us-east-1#logsV2:log-groups/log-group/dsc-ecs-dev/log-events/dsc$252Fdso-dsc-dev$252F9a4b97ed7f124da8b41b7af3d975cecc)) 
   2. submit **without** `collection-handle` ([step function execution](https://us-east-1.console.aws.amazon.com/states/home?region=us-east-1#/v2/executions/details/arn:aws:states:us-east-1:222053980223:execution:dso-dev:716fb0c4-ed31-49ae-b68d-e610d0af2240), [CloudWatch log stream](https://us-east-1.console.aws.amazon.com/cloudwatch/home?region=us-east-1#logsV2:log-groups/log-group/dsc-ecs-dev/log-events/dsc$252Fdso-dsc-dev$252Faef089ec2ec94957af18d7812c862176))
   3. submit **with** `collection-handle` ([step function execution](https://us-east-1.console.aws.amazon.com/states/home?region=us-east-1#/v2/executions/details/arn:aws:states:us-east-1:222053980223:execution:dso-dev:38558fb2-af07-4cc3-8777-fe171bd7ff22), [CloudWatch log stream](https://us-east-1.console.aws.amazon.com/cloudwatch/home?region=us-east-1#logsV2:log-groups/log-group/dsc-ecs-dev/log-events/dsc$252Fdso-dsc-dev$252F769272e2efee473bb5210ec2917b5ce0))
   4. ingest with DSS ([step function execution](https://us-east-1.console.aws.amazon.com/states/home?region=us-east-1#/v2/executions/details/arn:aws:states:us-east-1:222053980223:execution:dso-dev:604e5739-bdfe-4e25-bd84-e7b047fdc04f), [CloudWatch log stream](url))
   5. finalize ([step function execution](https://us-east-1.console.aws.amazon.com/states/home?region=us-east-1#/v2/executions/details/arn:aws:states:us-east-1:222053980223:execution:dso-dev:d973da54-9580-4368-994a-95b87278df82), [CloudWatch log stream](https://us-east-1.console.aws.amazon.com/cloudwatch/home?region=us-east-1#logsV2:log-groups/log-group/dsc-ecs-dev/log-events/dsc$252Fdso-dsc-dev$252F316e81df69fe4feda241ae38ee2290b3))

See ingested item at: https://mit-dspace.eks.prod.4science.cloud/handle/1721.1/164758

YES

NO

- https://mitlibraries.atlassian.net/browse/IN-1688

### Code review

### DIFF
--- a/dsc/cli.py
+++ b/dsc/cli.py
@@ -224,7 +224,6 @@ def sync(
     "-c",
     "--collection-handle",
     help="The handle of the DSpace collection to which the batch will be submitted",
-    required=True,
 )
 @click.option(
     "-e",
--- a/dsc/cli.py
+++ b/dsc/cli.py
@@ -1,6 +1,5 @@
 import datetime
 import logging
-import subprocess
 from time import perf_counter
 
 import click
@@ -9,6 +8,7 @@ from dsc.config import Config
 from dsc.db.models import ItemSubmissionDB
 from dsc.exceptions import BatchCreationFailedError
 from dsc.reports import CreateReport, FinalizeReport, SubmitReport
+from dsc.utils.aws.s3 import run_aws_cli_sync
 from dsc.workflows.base import Workflow
 
 logger = logging.getLogger(__name__)
@@ -212,49 +212,10 @@ def sync(
             "or set the S3_BUCKET_SYNC_SOURCE environment variable"
         )
 
-    logger.info(f"Syncing data from {source} to {destination}")
-
-    args = [
-        "aws",
-        "s3",
-        "sync",
-        source,
-        destination,
-        "--delete",
-        "--exclude",
-        "dspace_metadata/*",
-    ]
-
-    # add optional args
-    if dry_run:
-        args.append("--dryrun")
-
-    process = subprocess.Popen(  # noqa: S603
-        args, stdout=subprocess.PIPE, stderr=subprocess.PIPE, text=True
+    return_code = run_aws_cli_sync(
+        source, destination, exclude_patterns=["dspace_metadata/*"], dry_run=dry_run
     )
-
-    # log process output (stdout and stderr) in real-time
-    if process.stdout:
-        for line in process.stdout:
-            if line:
-                logger.info(line)
-    else:
-        logger.info("No changes detected in source, no sync required")
-
-    if process.stderr:
-        for line in process.stderr:
-            if line:
-                logger.error(line)
-
-    # wait for the process to complete
-    process.wait()
-    return_code = process.returncode
-
-    if return_code != 0:
-        logger.error(f"Failed to sync (exit code: {return_code})")
-        ctx.exit(return_code)  # exit with the same code as subprocess
-    else:
-        logger.info("Sync completed successfully")
+    ctx.exit(return_code)
 
 
 @main.command()
--- a/dsc/utils/aws/s3.py
+++ b/dsc/utils/aws/s3.py
@@ -24,7 +24,10 @@ class S3Client:
         source_file: str,
         destination_file: str,
     ) -> None:
-        """Update the key of the specified file to archive it from processing.
+        """Move an S3 object to another location.
+
+        Like the AWS CLI 'mv' command, this method copies the source object or file
+        to the specified destination and then deletes the source object or file.
 
         Args:
             source_file: S3 URI to source object to copy.
@@ -128,8 +131,8 @@ class S3Client:
 def run_aws_cli_sync(
     source: str,
     destination: str,
-    exclude_patterns: list[str] | None = None,
     *,
+    exclude_patterns: list[str] | None = None,
     dry_run: bool = False,
 ) -> int:
     logger.info(f"Syncing data from {source} to {destination}")
@@ -141,8 +144,7 @@ def run_aws_cli_sync(
     if exclude_patterns:
         for pattern in exclude_patterns:
             args.extend(["--exclude", pattern])
-
-    # displays the operations that would be performed without actually running
+    # only display operations that would be performed without execution
     if dry_run:
         args.append("--dryrun")
 
--- a/dsc/utils/aws/s3.py
+++ b/dsc/utils/aws/s3.py
@@ -1,6 +1,7 @@
 from __future__ import annotations
 
 import logging
+import subprocess
 from typing import TYPE_CHECKING
 
 import boto3
@@ -107,3 +108,53 @@ class S3Client:
                         continue
 
                     yield f"s3://{bucket}/{content["Key"]}"
+
+
+def run_aws_cli_sync(
+    source: str,
+    destination: str,
+    exclude_patterns: list[str] | None = None,
+    *,
+    dry_run: bool = False,
+) -> int:
+    logger.info(f"Syncing data from {source} to {destination}")
+
+    args = ["aws", "s3", "sync", source, destination, "--delete"]
+
+    # add optional args
+    # exclude all files or objects from the command that matches the specified patterns
+    if exclude_patterns:
+        for pattern in exclude_patterns:
+            args.extend(["--exclude", pattern])
+
+    # displays the operations that would be performed without actually running
+    if dry_run:
+        args.append("--dryrun")
+
+    process = subprocess.Popen(  # noqa: S603
+        args, stdout=subprocess.PIPE, stderr=subprocess.PIPE, text=True
+    )
+
+    # log process output (stdout and stderr) in real-time
+    if process.stdout:
+        for line in process.stdout:
+            if line:
+                logger.info(line)
+    else:
+        logger.info("No changes detected in source, no sync required")
+
+    if process.stderr:
+        for line in process.stderr:
+            if line:
+                logger.error(line)
+
+    # wait for the process to complete
+    process.wait()
+    return_code = process.returncode
+
+    if return_code != 0:
+        logger.error(f"Failed to sync (exit code: {return_code})")
+    else:
+        logger.info("Sync completed successfully")
+
+    return return_code
--- a/dsc/utils/aws/s3.py
+++ b/dsc/utils/aws/s3.py
@@ -3,6 +3,7 @@ from __future__ import annotations
 import logging
 import subprocess
 from typing import TYPE_CHECKING
+from urllib.parse import urlparse
 
 import boto3
 
@@ -18,25 +19,39 @@ class S3Client:
     def __init__(self) -> None:
         self.client = boto3.client("s3")
 
-    def archive_file_with_new_key(
-        self, bucket: str, key: str, archived_key_prefix: str
+    def move_file(
+        self,
+        source_file: str,
+        destination_file: str,
     ) -> None:
         """Update the key of the specified file to archive it from processing.
 
         Args:
-            bucket: The S3 bucket containing the files to be archived.
-            key: The key of the file to archive.
-            archived_key_prefix: The prefix to be applied to the archived file.
+            source_file: S3 URI to source object to copy.
+            destination_file: S3 URI to destination object.
         """
+        parsed_source_uri = urlparse(source_file, allow_fragments=False)
+        parsed_destination_uri = urlparse(destination_file, allow_fragments=False)
+
+        source_bucket, source_key = (
+            parsed_source_uri.netloc,
+            parsed_source_uri.path.lstrip("/"),
+        )
+        destination_bucket, destination_key = (
+            parsed_destination_uri.netloc,
+            parsed_destination_uri.path.lstrip("/"),
+        )
+
         self.client.copy_object(
-            Bucket=bucket,
-            CopySource=f"{bucket}/{key}",
-            Key=f"{archived_key_prefix}/{key}",
+            Bucket=destination_bucket,
+            CopySource=f"{source_bucket}/{source_key}",
+            Key=destination_key,
         )
         self.client.delete_object(
-            Bucket=bucket,
-            Key=key,
+            Bucket=source_bucket,
+            Key=source_key,
         )
+        logger.debug(f"Moved file from {source_file} to {destination_file}")
 
     def put_file(
         self,
--- a/dsc/workflows/base/workflow.py
+++ b/dsc/workflows/base/workflow.py
@@ -239,7 +239,7 @@ class Workflow(ABC):
         pass  # noqa: PIE790
 
     @final
-    def _create_batch_in_db(self, item_submissions: list[dict]) -> None:
+    def _create_batch_in_db(self, item_submissions: list[ItemSubmission]) -> None:
         """Write records for a batch of item submissions to DynamoDB.
 
         This method loops through the item submissions (init params)
@@ -247,8 +247,7 @@ class Workflow(ABC):
         method creates an instance of ItemSubmission and saves the
         record to DynamoDB.
         """
-        for item_submission_init_params in item_submissions:
-            item_submission = ItemSubmission.create(**item_submission_init_params)
+        for item_submission in item_submissions:
             item_submission.last_run_date = self.run_date
             item_submission.status = ItemSubmissionStatus.BATCH_CREATED
             item_submission.status_details = None
--- a/dsc/workflows/base/workflow.py
+++ b/dsc/workflows/base/workflow.py
@@ -101,7 +101,7 @@ class Workflow(ABC):
     """A base workflow class from which other workflow classes are derived."""
 
     workflow_name: str = "base"
-    submission_system: str = "DSpace@MIT"
+    submission_system: str = "IR-8"
 
     def __init__(self, batch_id: str) -> None:
         """Initialize base instance.
--- a/dsc/workflows/base/workflow.py
+++ b/dsc/workflows/base/workflow.py
@@ -253,8 +253,7 @@ class Workflow(ABC):
             item_submission.status_details = None
             item_submission.save()
 
-    @final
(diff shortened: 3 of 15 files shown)


pyproject.toml
```diff
@@ -10,28 +10,33 @@ requires-python = "~=3.12.0"
 dependencies = [
     "boto3>=1.42.30",
     "boto3-stubs[essential, ses]>=1.42.30",
+    "bs4>=0.0.2",
     "click>=8.3.1",
+    "dspace-rest-client==0.1.16",
```
ehanson8: Do we need to pin this?

dsc/workflows/opencourseware/workflow.py
```diff
@@ -120,11 +121,11 @@ def prepare_batch(
...
-                {
-                    "batch_id": self.batch_id,
-                    "item_identifier": item_metadata["item_identifier"],
-                    "workflow_name": self.workflow_name,
-                }
+                ItemSubmission(
+                    batch_id=self.batch_id,
+                    item_identifier=item_metadata["item_identifier"],
+                    workflow_name=self.workflow_name,
+                )
```
ehanson8, line 128: Smart fix, don't know why we didn't do this before!

dsc/utils/aws/s3.py
```diff
@@ -107,3 +108,53 @@ def files_iter(
...
+    # add optional args
+    # exclude all files or objects from the command that matches the specified patterns
+    if exclude_patterns:
+        for pattern in exclude_patterns:
+            args.extend(["--exclude", pattern])
+
+    # displays the operations that would be performed without actually running
+    if dry_run:
+        args.append("--dryrun")
+
```
ehanson8, line 150: Optional: another comment here to show the `subprocess` command is actually running rather than being a part of `# displays the operations that would be performed without actually running`

dsc/utils/aws/s3.py
```diff
@@ -107,3 +108,53 @@ def files_iter(
...
+    process = subprocess.Popen(  # noqa: S603
+        args, stdout=subprocess.PIPE, stderr=subprocess.PIPE, text=True
+    )
+
+    # log process output (stdout and stderr) in real-time
+    if process.stdout:
+        for line in process.stdout:
+            if line:
+                logger.info(line)
```
ehanson8, line 159: Use list comprehension here?

dsc/utils/aws/s3.py
```diff
@@ -107,3 +108,53 @@ def files_iter(
...
+    else:
+        logger.info("No changes detected in source, no sync required")
+
+    if process.stderr:
+        for line in process.stderr:
+            if line:
+                logger.error(line)
```
ehanson8, line 166: Same as above

dsc/cli.py
```diff
@@ -212,49 +212,10 @@ def sync(
...
-        "dspace_metadata/*",
-    ]
-
-    # add optional args
-    if dry_run:
-        args.append("--dryrun")
-
-    process = subprocess.Popen(  # noqa: S603
-        args, stdout=subprocess.PIPE, stderr=subprocess.PIPE, text=True
+    return_code = run_aws_cli_sync(
```
ehanson8, line 215: Beautifully slimple now!

dsc/utils/aws/s3.py
```diff
@@ -18,25 +19,39 @@ class S3Client:
...
         self.client = boto3.client("s3")
 
-    def archive_file_with_new_key(
-        self, bucket: str, key: str, archived_key_prefix: str
+    def move_file(
+        self,
+        source_file: str,
+        destination_file: str,
     ) -> None:
         """Update the key of the specified file to archive it from processing.
```
ehanson8: Great call making this generic but I would update the docstring which is tightly coupled to the previous method name

dsc/workflows/base/workflow.py
```diff
@@ -335,6 +340,19 @@ def submit_items(self, collection_handle: str) -> list:
...
+        """Get collection handle for an item submission.
+
+        This method is required for workflows where the collection handle for an item
+        must be derived dynamically and usually from the item metadata.
+
+        OPTIONAL override by workflow subclasses.
+        """
+        raise NotImplementedError(
+            f"The '{self.workflow_name}' workflow expects collection_handle"
+            "when calling submit_items()"
```
ehanson8, line 353: Seems good and I know you're in progress with more changes that will illuminate the actual use case but wondering if it should be saved until then. But I am OK if you'd just like to push this for now!

dsc/workflows/base/workflow.py
```diff
@@ -101,7 +101,7 @@ class Workflow(ABC):
     """A base workflow class from which other workflow classes are derived."""
 
     workflow_name: str = "base"
-    submission_system: str = "DSpace@MIT"
+    submission_system: str = "IR-8"
```
ehanson8, line 104: We may want to check with Sadie on whether the D8 instance is still going to be referred to as `DSpace@MIT` but it's easy to change now

ehanson8 (commented):
Excellent work! A few questions

dsc/workflows/opencourseware/workflow.py
ghukill: +1

dsc/utils/aws/s3.py
ghukill: I'd pushback on this here.  Given taht `stdout` is emitting log lines _as they happen_, I think there is value in logging them as they happen as well.  With list comprehension, you'd be waiting until the full sync was complete to see anything.

This feels aligned with how we exercise `stdout` / `stderr` iterators for running subprocesses in other projects.  Here is an example from browsertrix-harvester: https://github.com/MITLibraries/browsertrix-harvester/blob/main/harvester/crawl.py#L87-L99.

dsc/cli.py
ghukill: Agreed!

dsc/utils/aws/s3.py
ghukill: Agreed!  It's agnostic as to what kind of files it's moving.

dsc/workflows/base/workflow.py
```diff
@@ -253,8 +253,7 @@ def _create_batch_in_db(self, item_submissions: list[ItemSubmission]) -> None:
             item_submission.status_details = None
             item_submission.save()
 
-    @final
-    def submit_items(self, collection_handle: str) -> list:
+    def submit_items(self, collection_handle: str | None = None) -> list:
```
ghukill, line 256: Commenting here, but not a formal request for this PR.

I think this method might benefit from leaning on another like `_submit_item(...)`.  Then, this could be simplified as a `for` loop that orchestrates, but the private `_submit_item(...)` method has most of the business logic.

You'd end up with something more like this, with the `try` only attemping one statement:

```python
for item_submission in ItemSubmission.get_batch(self.batch_id):
    self.submission_summary["total"] += 1

    if not item_submission.ready_to_submit():
        self.submission_summary["skipped"] += 1
        continue

    try:
        items.append(
            self._submit_item(item_submission, collection_handle, batch_metadata)
        )

    except NotImplementedError:
        raise

    except Exception as exception:  # noqa: BLE001
        self.submission_summary["errors"] += 1
        item_submission.status = ItemSubmissionStatus.SUBMIT_FAILED
        item_submission.status_details = str(exception)
        item_submission.upsert_db()
```

What I feel like that refactor work does is expose how `collection_handle` and `batch_metadata` continue to feel kind of awkward.  

It would appear that for any given invocation of `submit_items()` they are shared across _all_ items, so why couldn't they be saved to the workflow instance? **or**, attached to the `ItemSubmission` instances that are getting passed around?

Feel free to ignore this for now, but I may reference it for the collection handle resolution if not provided in another comment.

dsc/workflows/base/workflow.py
ghukill: I'm in the same boat.  I'm good with this scaffolding  getting added, but I think either a PR comment or an inline comment of what that method  _might_ look like when implemented for a workflow would be helpful.  

For example: is it making an API request to DSpace?  if so, when we call it in `submit_items()` I feel like we may want to unearth it a bit from a tertiary `collection_handle or self._get_item_collection_handle()` call which potentially hides the fact it may be doing an expensive operation.

dsc/workflows/base/workflow.py
ghukill: I'm circling back to this now given my continued reading and [comment down below](https://github.com/MITLibraries/dspace-submission-composer/pull/225/changes/ddbe504a0918851258c3867b744b0f04c7cb9018#r3102069200).

I get how this works, but am finding there is some cognitive dissonance here.

Is it correct that for _most_ workflows, `collection_handle` is static / fixed / same for all items, and that's how it gets passed to `submit_items()`?  But for some -- e.g. digi theses -- it is not, and can theoretically vary per item?

If so, I think I'd double down and _always_ use a dedicated method for getting the collection handle.  And, make _all_ workflows implement it.  Maybe 90% of them just return a single, hardcoded, static string, but then there is parity on how a single item gets its collection handle: it calls `get_collection_handle(<ItemSubmission>)` or something.

This still leaves the awkward `batch_metadata` in my comment above about refactoring to a private `_submit_item()`, but it knocks out the `collection_handle` passing weirdness!

ghukill (commented):
These are some thoughtful and targeted changes, nice work!

Please see my comments about considering a new private `_submit_item()` (singular), how this exposes `collection_handle` as kind of awkward, and a possible doubling down on the getting of collection handles as a first class action that might deserve a method that is _always_ called, not just when `collection_handle = None`.

pyproject.toml
jonavellecuerdo: Good question. I'm not sure! I will remove `dspace-rest-client` and `bs4` for now as they are not used in any of the updates introduced in this PR.

dsc/utils/aws/s3.py
jonavellecuerdo: To clarify, [this line](https://github.com/MITLibraries/dspace-submission-composer/blob/IN-1626-digitized-theses-prereqs/dsc/utils/aws/s3.py#L149-L151) will always be executed; it is only a matter of whether optional args are passed into the `subprocess.Popen(args)` call.

Updated to: 

```python
# add optional args
# exclude all files or objects from the command that matches the specified patterns
if exclude_patterns:
   for pattern in exclude_patterns:
      args.extend(["--exclude", pattern])
# only display operations that would be performed without execution
if dry_run:
   args.append("--dryrun")
```

Removed whitespace to indicate how `dry_run` and `exclude_patterns` are optional args.

dsc/utils/aws/s3.py
jonavellecuerdo: I agree that we should log lines as they are emitted from `stdout`. Since we are not using the log output, I do not see a need to store the logged messages in a list (as would be done by list comprehension).

dsc/utils/aws/s3.py
jonavellecuerdo: See respone above: https://github.com/MITLibraries/dspace-submission-composer/pull/225#discussion_r3126131079

dsc/workflows/base/workflow.py
jonavellecuerdo: @ghukill Hmm, I'm not sure about adding pseudo-code for what the method might look like as it can vary by workflow. 🤔 For its first use case with Digitized Theses, we are using some field in the metadata + regex to map each item submission to a known record collection (i.e., does not use the API). In that case, I think the docstring for the defined method in `DigitizedTheses` workflow will contain more detail. 

Is making this slight update in the docstring sufficient:

> This method is required for workflows where the collection handle for an item
        must be derived dynamically based on the provided item metadata.

dsc/utils/aws/s3.py
jonavellecuerdo: Updated the description:

```python
"""Move an S3 object to another location.

Like the AWS CLI 'mv' command, this method copies the source object or file
to the specified destination and then deletes the source object or file.

...
```

Thanks for the callout!

dsc/workflows/base/workflow.py
jonavellecuerdo: To recap, the `collection_handle` is passed into DSC ad follows:
* Users can provide a `collection_handle` in the input payload for `submit` command when invoking the Step Function.
* The `collection_handle` is set for the DSC `submit` CLI command.
* The `collection_handle` is passed to `Workflow.submit_items()` call.
* [The `collection_handle` is set on the `ItemSubmission` object.](https://github.com/MITLibraries/dspace-submission-composer/blob/IN-1626-digitized-theses-prereqs/dsc/workflows/base/workflow.py#L301-L303)

---
In my initial passes at defining this method, I was planning for the following signature: `_get_item_collection_handle(item_metadata: dict)`, rather than accepting an `ItemSubmission` instance (for context, item metadata is not saved to the `ItemSubmission` object and is thus not present when the object is loaded from DynamoDB). Thus, item metadata has to come from _somewhere_, and currently it is from [this line](https://github.com/MITLibraries/dspace-submission-composer/blob/IN-1626-digitized-theses-prereqs/dsc/workflows/base/workflow.py#L272-L275) (which is less than ideal).

Continuing the assumption that item metadata is not saved to the `ItemSubmission` object, if we were to define the method for all workflows, a workflow like `opencourseware` or `sccs` would look like this:

```python
class Workflow:
   def submit_items(self, collection_handle: str):
      get_item_collection_handle(collection_handle)

   # where
   def get_item_collection_handle(collection_handle: str): 
      return collection_handle
```

The sample code above feels redundant/somewhat of an anti-pattern. 🤔 For this reason, I proposed:

```python
class Workflow:
   def submit_items(self, collection_handle: str | None = None): 
      item_submission.collection_handle = (
                    collection_handle or self._get_item_collection_handle(<item metadata accessed from somewhere>)
                )
```

---

I hope this provides some context into the updates proposed here! Lastly, I selected the name `_get_item_collection_handle` in hopes it would imply how, if the method is called, this collection handle is at the _item-level_ and not intended to be used across all the item submissions for a given batch. 

Let me know what you think about keeping the derivation as is. Fully acknowledging that we still should tackle how we access item metadata in a follow-up PR!

Tagging @ehanson8 to request additional feedback and thoughts, too.

dsc/utils/aws/s3.py
ehanson8: Got it!

dsc/utils/aws/s3.py
ehanson8: That's fine!

dsc/utils/aws/s3.py
ehanson8: Perfect!

dsc/workflows/base/workflow.py
ghukill: Yeah, happy to retract this comment!  I'm not 100% sure what I was thinking originally.  I'll continue to look at the other collection handle comments / discussions, maybe there is correlation to this `raise NotImplementedError`, but perhaps not.

dsc/workflows/base/workflow.py
ghukill: @jonavellecuerdo - you're definitely much closer to the codebase, so I say go for it if it's feeling right still.  I can always dig back in post PR and, if something still feels off, propose a small new PR at that time.  Thanks for hearing me out!

ghukill (approved):

ehanson8 (approved):

dsc/workflows/base/workflow.py
jonavellecuerdo: I will merge as-is for now but continue to ponder in next PR!

PR #226, opened 2026-04-06
Add metrics client

ehanson8:
This PR creates a `MetricsClient` class that will eventually be added to our template repos and provides an opportunity to refine our conventions for publishing AWS metrics. 

This is not easily to manually test but you can review this [Cloudwatch query](https://us-east-1.console.aws.amazon.com/cloudwatch/home?region=us-east-1#metricsV2?graph=~(metrics~(~(~'dso~'ingest_error~'application~'dsc~'workflow_name~'sccs~(id~'m1))~(~'.~'item_submitted~'.~'.~'.~'.~(id~'m2))~(~'.~'ingested_item~'.~'.~'.~'.~(id~'m3)))~sparkline~false~view~'timeSeries~stacked~false~region~'us-east-1~stat~'Sum~period~3600~start~'-PT168H~end~'P0D)&query=~'*7bdso*2capplication*2cworkflow_name*7d) to see the metrics that have been produced by this code over the past week

NO

NO

- NA

### Code review

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -1,29 +1,28 @@
 default_language_version:
-  python: python3.12 # set for project python version
+  python: python3.12
+default_stages:
+  - pre-push
 repos:
   - repo: local
     hooks:
-      - id: black-apply
-        name: black-apply
-        entry: uv run black
+      - id: ruff-format
+        name: ruff-format
+        entry: uv run ruff format --diff
         language: system
         pass_filenames: true
-        types: ["python"]
+        types: [ "python" ]
+
       - id: mypy
         name: mypy
         entry: uv run mypy
         language: system
         pass_filenames: true
-        types: ["python"]
-        exclude: "tests/"
-      - id: ruff-apply
-        name: ruff-apply
-        entry: uv run ruff check --fix
+        types: [ "python" ]
+        exclude: "(tests/|output/|migrations/)"
+
+      - id: ruff-check
+        name: ruff-check
+        entry: uv run ruff check
         language: system
         pass_filenames: true
-        types: ["python"]
-      - id: pip-audit
-        name: pip-audit
-        entry: uv run pip-audit
-        language: system
-        pass_filenames: false
+        types: [ "python" ]
--- a/dsc/utils/aws/metrics.py
+++ b/dsc/utils/aws/metrics.py
@@ -1,0 +1,198 @@
+"""AWS CloudWatch metrics client for workflow submission tracking."""
+
+from __future__ import annotations
+
+import logging
+
+import boto3
+
+from dsc.config import METRICS, METRICS_NAMESPACE
+
+logger = logging.getLogger(__name__)
+
+
+UNIT_VALUES = frozenset(
+    [
+        "Seconds",
+        "Microseconds",
+        "Milliseconds",
+        "Bytes",
+        "Kilobytes",
+        "Megabytes",
+        "Gigabytes",
+        "Terabytes",
+        "Bits",
+        "Kilobits",
+        "Megabits",
+        "Gigabits",
+        "Terabits",
+        "Percent",
+        "Count",
+        "Bytes/Second",
+        "Kilobytes/Second",
+        "Megabytes/Second",
+        "Gigabytes/Second",
+        "Terabytes/Second",
+        "Bits/Second",
+        "Kilobits/Second",
+        "Megabits/Second",
+        "Gigabits/Second",
+        "Terabits/Second",
+        "Count/Second",
+    ]
+)
+
+
+class MetricsClient:
+    """A simple client to record metrics to AWS CloudWatch."""
+
+    def __init__(self) -> None:
+        """Initialize the MetricsClient."""
+        self.cloudwatch = boto3.client("cloudwatch")
+        self.batch_metrics: list[dict] = []
+
+    def publish_single_metric(
+        self,
+        metric_name: str,
+        value: int,
+        unit: str,
+        metric_dimensions: dict[str, str] | None = None,
+    ) -> None:
+        """Publish a single metric to CloudWatch.
+
+        Args:
+            metric_name: The name of the metric to publish.
+            value: The value of the metric.
+            unit: The unit of the metric.
+            metric_dimensions: Optional dictionary of dimension names and values.
+
+        Raises:
+            ValueError: If unit is invalid.
+        """
+        metric_data = self._validate_and_build_metric_data(
+            metric_name, value, unit, metric_dimensions
+        )
+        self._push_metric_data([metric_data])
+
+    def _validate_and_build_metric_data(
+        self,
+        metric_name: str,
+        value: int,
+        unit: str,
+        metric_dimensions: dict[str, str] | None = None,
+    ) -> dict:
+        """Validate and build a metric data dictionary for CloudWatch.
+
+        Args:
+            metric_name: The name of the metric.
+            value: The value of the metric.
+            unit: The unit of the metric.
+            metric_dimensions: Optional dictionary of dimension names and values.
+
+        Returns:
+            A metric data dictionary formatted for CloudWatch.
+        """
+        self._approved_metric(metric_name)
+        self._validate_unit(unit)
+        dimensions = [
+            {"Name": name, "Value": dim_value}
+            for name, dim_value in (
+                metric_dimensions.items() if metric_dimensions else []
+            )
+        ]
+        return {
+            "MetricName": metric_name,
+            "Value": value,
+            "Unit": unit,
+            "Dimensions": dimensions,
+        }
+
+    def _approved_metric(self, metric_name: str) -> bool:
+        """Check if a metric name is in the approved list of metrics for the application.
+
+        Args:
+            metric_name: The name of the metric to check.
+        """
+        if metric_name not in METRICS:
+            raise ValueError(
+                f"Metric name '{metric_name}' is not in the approved list of metrics: "
+                f"{', '.join(METRICS)}"
+            )
+        return True
+
+    def _validate_unit(self, unit: str) -> None:
+        """Validate that metric unit is allowed by AWS CloudWatch.
+
+        Args:
+            unit: The unit to validate.
+
+        Raises:
+            ValueError: If unit is not allowed by AWS CloudWatch.
+        """
+        if unit not in UNIT_VALUES:
+            raise ValueError(
+                f"Invalid unit '{unit}'. Must be one of: {', '.join(UNIT_VALUES)}"
+            )
+
+    def _push_metric_data(self, metric_data: list[dict]) -> None:
+        """Push metric data to CloudWatch.
+
+        Args:
+            metric_data: List of metric dictionaries to push.
+        """
+        try:
+            self.cloudwatch.put_metric_data(
+                Namespace=METRICS_NAMESPACE, MetricData=metric_data
+            )
+            logger.info(f"Published metric with {metric_data} to CloudWatch.")
+        except Exception:
+            logger.exception(
+                f"Failed to publish metric with {metric_data} to CloudWatch."
+            )
+
+    def add_metric_to_batch(
+        self,
+        metric_name: str,
+        value: int,
+        unit: str,
+        metric_dimensions: dict[str, str] | None = None,
+    ) -> None:
+        """Add a metric to the batch for later publishing.
+
+        Args:
+            metric_name: The name of the metric.
+            value: The value of the metric.
+            unit: The unit of the metric.
+            metric_dimensions: Optional dictionary of dimension names and values.
+
+        Raises:
+            ValueError: If unit is invalid.
+        """
+        metric_data = self._validate_and_build_metric_data(
+            metric_name, value, unit, metric_dimensions
+        )
+        self.batch_metrics.append(metric_data)
+
+    def publish_batch_metrics(self, batch_size: int = 20) -> None:
+        """Publish all accumulated batch metrics to CloudWatch.
+
+        Raises:
+            ValueError: If any metric has an invalid unit or missing required fields.
+        """
+        if not self.batch_metrics:
+            logger.info("No metrics to publish.")
+            return
+
+        # Validate all metrics before publishing
+        for metric in self.batch_metrics:
+            if not all(key in metric for key in ["MetricName", "Value", "Unit"]):
+                raise ValueError(
+                    f"Each metric must contain 'MetricName', 'Value', and 'Unit'. "
+                    f"Invalid metric: {metric}"
+                )
+            self._approved_metric(metric["MetricName"])
+            self._validate_unit(metric["Unit"])
+
+        for x in range(0, len(self.batch_metrics), batch_size):
+            self._push_metric_data(self.batch_metrics[x : x + batch_size])
+        self.batch_metrics.clear()
--- a/dsc/utils/aws/metrics.py
+++ b/dsc/utils/aws/metrics.py
@@ -116,9 +116,8 @@ class MetricsClient:
             ValueError: If unit is not allowed by AWS CloudWatch.
         """
         if unit not in UNIT_VALUES:
-            raise ValueError(
-                f"Invalid unit '{unit}'. Must be one of: {', '.join(sorted(UNIT_VALUES))}"
-            )
+            allowed_units = ", ".join(sorted(UNIT_VALUES))
+            raise ValueError(f"Invalid unit '{unit}'. Must be one of: {allowed_units}")
         return True
 
     def _publish_metrics(self, metrics: list[Metric]) -> None:
--- a/dsc/utils/aws/metrics.py
+++ b/dsc/utils/aws/metrics.py
@@ -56,7 +56,9 @@ class Metric:
 class MetricsClient:
     """A simple client to record metrics to AWS CloudWatch."""
 
-    def __init__(self, namespace: str, allowed_metrics: set[str] | None = None) -> None:
+    def __init__(
+        self, namespace: str | None = None, allowed_metrics: set[str] | None = None
+    ) -> None:
         """Initialize the MetricsClient."""
         self.namespace = namespace
         self.allowed_metrics: set[str] | None = allowed_metrics
--- a/dsc/utils/aws/metrics.py
+++ b/dsc/utils/aws/metrics.py
@@ -3,10 +3,11 @@
 from __future__ import annotations
 
 import logging
+from dataclasses import dataclass
 
 import boto3
 
-from dsc.config import METRICS, METRICS_NAMESPACE
+from dsc.config import METRICS_NAMESPACE
 
 logger = logging.getLogger(__name__)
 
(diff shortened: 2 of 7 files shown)


dsc/utils/aws/metrics.py
```diff
@@ -0,0 +1,198 @@
...
+        """Publish all accumulated batch metrics to CloudWatch.
+
+        Raises:
+            ValueError: If any metric has an invalid unit or missing required fields.
+        """
+        if not self.batch_metrics:
+            logger.info("No metrics to publish.")
+            return
+
+        # Validate all metrics before publishing
```
ehanson8: While this feels slightly weird to re-validate as they were validated in `add_metric_to_batch`, there is the potential for something to go haywire in between when they are added and when they are published. A cleaner solution would be much appreciated!

dsc/utils/aws/metrics.py
```diff
@@ -0,0 +1,198 @@
...
+                raise ValueError(
+                    f"Each metric must contain 'MetricName', 'Value', and 'Unit'. "
+                    f"Invalid metric: {metric}"
+                )
+            self._approved_metric(metric["MetricName"])
+            self._validate_unit(metric["Unit"])
+
+        for x in range(0, len(self.batch_metrics), batch_size):
+            self._push_metric_data(self.batch_metrics[x : x + batch_size])
+        self.batch_metrics.clear()
```
ehanson8, line 236: Forgoing unit test for now, but unit tests will certainly be included when the `MetricsClient` is added to the template repos.

dsc/utils/aws/metrics.py
```diff
@@ -0,0 +1,198 @@
...
+    ]
+)
+
+
+class MetricsClient:
+    """A simple client to record metrics to AWS CloudWatch."""
+
+    def __init__(self) -> None:
+        """Initialize the MetricsClient."""
+        self.cloudwatch = boto3.client("cloudwatch")
```
ghukill: Feeling like maybe we should make this a private property, like `self._cloudwatch`?  The _first_ thing I do with a new class/client like this is fire it up in a console and see what's available, what's surprising, etc., and this was a little surprising!

Image

dsc/utils/aws/metrics.py
```diff
@@ -0,0 +1,198 @@
...
+            "Dimensions": dimensions,
+        }
+
+    def _approved_metric(self, metric_name: str) -> bool:
+        """Check if a metric name is in the approved list of metrics for the application.
+
+        Args:
+            metric_name: The name of the metric to check.
+        """
+        if metric_name not in METRICS:
```
ghukill: So first of all, I really like the ability to validate the metric name itself.  But the ergonomics feel just a little off to me, given that it's hardcoded to expect a `METRICS` variable in scope somehow.  

I think this tightly couples the client to the nature of the project's config.

What if the class instantiation would support an `allowed_metrics: list[str] | None = None`?  

In code, if `None`, assume the user doesn't care about metric name validation.  But if set, validate against that list.

In code, I think it'd be as trivial as instantiating like this:

```python
from dsc.config import ALLOWED_METRICS

metrics_client = MetricsClient(allowed_metrics=ALLOWED_METRICS)
```

ghukill: I may find a place for it elsewhere, but I think the same would apply to `NAMESPACE` which has the same kind of tight coupling.

ghukill: To be fair: this is a test implementation that will get ported to the templates... but insofar as _that_ implementation will need to be a bit more application structure agnostic, so too must this.

ghukill: Lastly, see my actual PR comment review.  I'm proposing the use of a `Metric` dataclass, but it can't really validate the metric name if we go this route where it's not a magic value in scope, but a value optionally passed during client init.  

As such, _this_ comment thread may suggest that `Metric` self-validation is not a great approach, and it should remain here in the client.  I will ammend  that comment before sending.

ghukill (commented):
Before I jump too far into a review, as I was looking at the overall structure, I had a thought/question that might have some rippling effects.

What if we had a `dataclass` like `Metric`?

```python
@dataclass
class Metric:
    metric_name: str
    value: int | float | str # maybe others?
    unit: str
    metric_dimensions: dict[str, str] | None = None
```

If this was defined in the same module, that dataclass could handle validation during initialization, thereby removing the need to validate anywhere else in the class.  I'm not crazy about the idea, but I think you can use `@dataclass(frozen=True)` and a dataclass instance can't be modified.  If so, then you avoid that double validation approach to guard against a once-valid-but-maybe-not-anymore situation.

_UPDATE: leaving the above thinking about validation, but see also [this comment](https://github.com/MITLibraries/dspace-submission-composer/pull/226/changes#r3045502667).  It feels correct to init the client with allowed metrics, therefore it doesn't make sense for this proposed `Metric` dataclass to validate itself._

It would also begin to chip away at this repeating block for most methods (not to mention the docstrings):

```python
name: str,
value: int,
unit: str,
metric_dimensions: dict[str, str] | None = None,
```

If we had something like that, I could imagine method signatures along the lines of:

```python
def publish_metric(self, metric:Metric): ...

def add_metric_to_batch(self, metric:Metric): ...

def publish_metrics_batch(self): ...
```

To actually use inline in code, could just init as part of the publish:

```python
client.publish_metric(Metric(name="foo", value=42, unit="seconds"))
```

Whattya think?  Even if you don't go quite this far, and only want to pass around a new `Metric` dataclass _internally_ between methods, I do still think it could be a good place for validation and defining those metric properties only once.

dsc/utils/aws/metrics.py
ghukill: My first comment review + inline comments touches on this!

dsc/utils/aws/metrics.py
ehanson8: Done!

ehanson8:
Pushed a new commit based on @ghukill's comments and here's a [new query](https://us-east-1.console.aws.amazon.com/cloudwatch/home?region=us-east-1#metricsV2?graph=~(metrics~(~(~'dso~'ingest_error~'application~'dsc~'workflow_name~'sccs~(id~'m1))~(~'.~'item_submitted~'.~'.~'.~'.~(id~'m2))~(~'.~'ingested_item~'.~'.~'.~'.~(id~'m3)))~sparkline~false~view~'timeSeries~stacked~false~region~'us-east-1~stat~'Sum~period~3600~start~'-PT168H~end~'P0D)&query=~'*7bdso*2capplication*2cworkflow_name*7d) showing the results of the updated code

dsc/utils/aws/metrics.py
```diff
@@ -0,0 +1,190 @@
...
+            batch_size: Number of metrics to publish in each batch.
+        """
+        if not self.batch_metrics:
+            logger.info("No metrics to publish.")
+            return
+
+        try:
+            # Re-validate all metrics before publishing to CloudWatch
+            for metric in self.batch_metrics:
+                self._validate_metric(metric)
```
ghukill: Though the performance hit is almost nill, I'm not convinced we need this double validation here.  If users use `.add_metric_to_batch()` then it's already validated.  Yes, they could modify it as it sits in the queue before publishing... but that seems really unlikely and would have to be almost intentional.

For simplicity, I'd vote to drop this double validation.

I feel like also this sidesteps a possible bug.  For a batch of metrics `[m1, m2, m3]` queued for publishing, if `m2` fails, it would throw a `ValueError` of some kind, wouldn't the `finally` block clear the queue and we'd lose the other good metrics?

dsc/utils/aws/metrics.py
```diff
@@ -0,0 +1,190 @@
...
+        self._cloudwatch = boto3.client("cloudwatch")
+        self.batch_metrics: list[Metric] = []
+
+    def publish_single_metric(
+        self,
+        metric: Metric,
+    ) -> None:
+        """Publish a single metric to CloudWatch."""
+        self._validate_metric(metric)
+        self._push_metric_data([metric])
```
ghukill: I like this re-suse of `_push_metric_data()`, just passing a single item in a list!

dsc/utils/aws/metrics.py
```diff
@@ -0,0 +1,190 @@
...
+    """A simple client to record metrics to AWS CloudWatch."""
+
+    def __init__(self, allowed_metrics: set[str] | None = None) -> None:
+        """Initialize the MetricsClient."""
+        self.namespace = METRICS_NAMESPACE
+        self.allowed_metrics: set[str] | None = allowed_metrics
+        self._cloudwatch = boto3.client("cloudwatch")
+        self.batch_metrics: list[Metric] = []
+
+    def publish_single_metric(
```
ghukill: What about just `publish_metric()`?  Nice symmetry with `_validate_metric()`.

dsc/utils/aws/metrics.py
```diff
@@ -0,0 +1,190 @@
...
+        Raises:
+            ValueError: If unit is not allowed by AWS CloudWatch.
+        """
+        if unit not in UNIT_VALUES:
+            raise ValueError(
+                f"Invalid unit '{unit}'. Must be one of: {', '.join(UNIT_VALUES)}"
+            )
+        return True
+
+    def _push_metric_data(self, metrics: list[Metric]) -> None:
```
ghukill: What about `_publish_metrics()`?  The plural is important, as is that it's a private method.

dsc/utils/aws/metrics.py
```diff
@@ -0,0 +1,190 @@
...
+    def add_metric_to_batch(self, metric: Metric) -> None:
+        """Add a metric to the batch queue.
+
+        Args:
+            metric: The metric to add to the batch.
+        """
+        self._validate_metric(metric)
+        self.batch_metrics.append(metric)
+
+    def publish_batch_metrics(self, batch_size: int = 20) -> None:
```
ghukill: What about `publish_metrics_batch()`.  Very subjective, but feels like it has more symmetry with other methods.  Pullling all the public methods here:

- `publish_metric()` = single metric, right now
- `add_metric_to_batch()` = queue a metric for later
- `publish_metrics_batch()` = publishes metrics and clears queue

dsc/utils/aws/metrics.py
```diff
@@ -0,0 +1,190 @@
...
+                f"Published {len(metrics)} metric(s) to CloudWatch namespace "
+                f"'{self.namespace}'."
+            )
+        except Exception:
+            logger.exception(
+                f"Failed to publish {len(metrics)} metric(s) to CloudWatch: "
+            )
+            raise
+
+    def add_metric_to_batch(self, metric: Metric) -> None:
```
ghukill: What about supporting both a single metric or a list of metrics? See example in PR comment.

If so, would likely want to rename to `add_metrics_to_batch()` (plural).

ghukill (commented):
Starting with a disclaimer that, as requested, going _very_ hard into nit picking on this PR with the hope / expectation that we could drop this `metrics.py` basically as-is into the template repositories and use elsewhere.

My inlined comments about method names and removing double validation are somewhat subjective and optional.

But there is a bigger piece around metrics **namespace** that I was having trouble finding a single place to inline comment.

I think that we should avoid namespace getting set by a `METRICS_NAMESPACE` imported into the metrics module.  The tightly couples the config for that app to expose a `METRICS_NAMESPACE` constant in the config file, not to mention just _requiring_ a config file (yes most have, but maybe not all do or will).

Additionally, we might find that a single application wants to publish metrics to multiple namespaces.

For these reasons, I might propose an alternate pattern:

1. _Optionally_ init `MetricsClient` with a namespace that will get set for metrics during publishing.  Consider supporting an env var default like `AWS_METRICS_NAMESPACE`.
2. _Optionally_ set `namespace` in the new `Metric` dataclass that will get used for that metric
3. A namespace in a single `Metric` instance will override a default namespace set for the client.
4. During metric validation, an error is thrown if _neither_ set a namespace for the client.

This would give applications a lot of flexibility of how they construct metrics an the metrics client itself.  

For example, in this DSC app, any given file or module could do:

```python
from dsc.config import Config
from dsc.utils.aws.metrics import MetricsClient

CONFIG = Config()

metrics_client = MetricsClient(
    namespace=CONFIG.metrics_namespace,
    allowed_metrics=CONFIG.allowed_metrics
)
```

Note that this assume the `Config` class has properties for these values instead of free floating constants in the `config.py` module.  But of course, either way works.  

Or, maybe the `config.py` has a metrics client factcory helper if lots of files need a metrics client:

```python
from dsc.utils.aws.metrics import MetricsClient

def get_metrics_client() -> MetricsClient:
    return MetricsClient(
        namespace=METRICS_NAMESPACE,
        allowed_metrics=ALLOWED_METRICS
    )
```

Note that in this example, the module level constants are used as currently they are set.  The flexibility is kind of key to me, in whatever state / conventions the app uses.

Then could use anywhere in the app:
```python
from dsc.config import get_metrics_client

metrics_client = get_metrics_client()
```

Lastly, an app may opt to publish to multiple namespaces.  For that app, a default namespace wouldn't be set on the client itself.  They would always prepare metrics with the namespace baked in:

```python
m1 = Metric(
    namespace="timdex",
    # rest of fields...
)

m2 = Metric(
    namespace="use",
    # rest of fields...
)

metrics_client.add_metrics_to_batch([m1, m2])
```

dsc/utils/aws/metrics.py
ehanson8: Good point, I'll remove!

dsc/utils/aws/metrics.py
ehanson8: I was trying to avoid potential confusion with `publish_metric` but the fact that it's private is probably enough to differentiate them

ghukill: And the plural "s".  Maybe it's a matter of taste, but I like APIs that use that convention.  It suggests that it's conceptually really similar, just "one" vs "many".

dsc/utils/aws/metrics.py
ehanson8: Agreed!

dsc/utils/aws/metrics.py
ehanson8: Good call!

ehanson8:
> For these reasons, I might propose an alternate pattern:
> 
> 1. _Optionally_ init `MetricsClient` with a namespace that will get set for metrics during publishing.  Consider supporting an env var default like `AWS_METRICS_NAMESPACE`.
> 2. _Optionally_ set `namespace` in the new `Metric` dataclass that will get used for that metric
> 3. A namespace in a single `Metric` instance will override a default namespace set for the client.
> 4. During metric validation, an error is thrown if _neither_ set a namespace for the client.

A great recommendation, working on this

ehanson8:
@ghukill Pushed a new commit! And the most recent [Cloudwatch query](https://us-east-1.console.aws.amazon.com/cloudwatch/home?region=us-east-1#metricsV2?graph=~(metrics~(~(~'dso~'ingest_error~'application~'dsc~'workflow_name~'sccs~(id~'m1))~(~'.~'item_submitted~'.~'.~'.~'.~(id~'m2))~(~'.~'ingested_item~'.~'.~'.~'.~(id~'m3)))~sparkline~false~view~'timeSeries~stacked~false~region~'us-east-1~stat~'Sum~period~3600~start~'-PT336H~end~'P0D)&query=~'*7bdso*2capplication*2cworkflow_name*7d)

dsc/utils/aws/metrics.py
```diff
@@ -52,25 +50,26 @@ class Metric:
...
 
 
 class MetricsClient:
     """A simple client to record metrics to AWS CloudWatch."""
 
-    def __init__(self, allowed_metrics: set[str] | None = None) -> None:
+    def __init__(self, namespace: str, allowed_metrics: set[str] | None = None) -> None:
         """Initialize the MetricsClient."""
-        self.namespace = METRICS_NAMESPACE
+        self.namespace = namespace
```
ghukill: If we're going to allow an individual `Metric` instance to define a namespace, I think that implies that `namespace` should be optional here.  This is why you need this validation,

> During metric validation, an error is thrown if neither set a namespace for the client.

which I do see in the new `_validate_namespace` method!  So think this just needs tso be `namespace: str | None`.

ghukill (approved):
I've requested a change, but approving as well.  If you feel the change is not correct, please feel free to re-request review from me.  

That little change aside, looking great!  Thanks for the all the discussion on this.  Feels like something we could drop right into the templates now.

dsc/utils/aws/metrics.py
ehanson8: Good catch, I'll update this!

dsc/utils/aws/metrics.py
```diff
@@ -0,0 +1,197 @@
...
+        self.batch_metrics: list[Metric] = []
+
+    def publish_metric(
+        self,
+        metric: Metric,
+    ) -> None:
+        """Publish a single metric to CloudWatch."""
+        self._validate_metric(metric)
+        self._publish_metrics([metric])
+
```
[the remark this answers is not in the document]
ehanson8: This will come in a later PR when the client is added the template repos

dsc/workflows/base/workflow.py
```diff
@@ -325,6 +332,15 @@ def submit_items(self, collection_handle: str | None = None) -> list:
...
                 item_submission.submit_attempts += 1
                 item_submission.upsert_db()
+
+                self.metrics_client.publish_metric(
+                    Metric(
+                        name="item_submitted",
+                        value=1,
+                        unit="Count",
+                        dimensions=self.metrics_dimensions,
+                    )
```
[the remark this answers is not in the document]
ehanson8: We're not expecting high-volume with this app so we'll proceed with the less-efficient option for now

dsc/utils/aws/metrics.py
```diff
@@ -0,0 +1,197 @@
...
+    def _publish_metrics(self, metrics: list[Metric]) -> None:
+        """Push metrics to CloudWatch.
+
+        Args:
+            metrics: List of metric instances to push.
+        """
+        if not metrics:
+            logger.info("No metrics to publish.")
+            return
+        try:
```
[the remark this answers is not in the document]
ehanson8: Discovered the AWS metric limit has actually been updated to 1000, implemented a defensive chunking approach

.pre-commit-config.yaml
```diff
@@ -1,29 +1,28 @@
...
         language: system
         pass_filenames: true
-        types: ["python"]
-      - id: pip-audit
-        name: pip-audit
-        entry: uv run pip-audit
-        language: system
-        pass_filenames: false
\ No newline at end of file
+        types: [ "python" ]
```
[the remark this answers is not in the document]
ehanson8: We still plan to manually run this check even if it's not a part of the pre-commit hook

dsc/utils/aws/metrics.py
```diff
@@ -0,0 +1,237 @@
...
+            for x in range(0, len(self.batch_metrics), batch_size):
+                batch = self.batch_metrics[x : x + batch_size]
+                self._publish_metrics(batch)
+        except Exception:
+            # Keep only the unpublished metrics (starting from the failed batch)
+            self.batch_metrics = self.batch_metrics[x:]
+            raise
+
+        # Clear only if all batches published successfully
+        self.batch_metrics.clear()
```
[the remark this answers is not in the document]
ehanson8: This seems like an unlikely risk so we will skip this check

dsc/utils/aws/metrics.py
```diff
@@ -0,0 +1,236 @@
...
+        """Check if a metric name is in the allowed list of metrics for the application.
+
+        Args:
+            metric_name: The name of the metric to check.
+        """
+        if self.allowed_metrics and metric_name not in self.allowed_metrics:
+            raise ValueError(
+                f"Metric name '{metric_name}' is not in the allowed list of metrics: "
+                f"{', '.join(self.allowed_metrics)}"
+            )
```
jonavellecuerdo, line 106: So if `self.allowed_metrics is None`, `_allowed_metric()` also returns True?

dsc/workflows/base/workflow.py
```diff
@@ -328,6 +335,20 @@ def submit_items(self, collection_handle: str | None = None) -> list:
...
+                            value=1,
+                            unit="Count",
+                            dimensions=self.metrics_dimensions,
+                        )
+                    )
+                except Exception:
+                    logger.exception(
+                        f"Failed to publish item submitted metric for item "
+                        f"{item_identifier}"
+                    )
```
jonavellecuerdo: Commenting here but applies to all the added `try-except` blocks with calls to `self.metrics_client.publish_metric(...)`. These try-except blocks are very similar, with the exception of `Metric.name` and the message in the logged exception.

Can you create a new private sub-method in workflow that encapsulates the try-except block, with params for the metric name? 🤔

jonavellecuerdo (changes requested):
This is looking good! Had a question regarding metric "namespaces" (i.e., unresolved comment raised by Copilot) and a change request to encapsulate metrics `try-except` blocks into a private sub-method.

dsc/utils/aws/metrics.py
ehanson8: Yes, `allowed_metrics` is optional so this method shouldn't block anything if `allowed_metrics` is `None`

dsc/utils/aws/metrics.py
```diff
@@ -0,0 +1,197 @@
...
+                metric_data.append(metric_dict)
+
+            self._cloudwatch.put_metric_data(
+                Namespace=metric.namespace or self.namespace,
+                MetricData=metric_data,
+            )
+            logger.info(
+                f"Published {len(metrics)} metric(s) to CloudWatch namespace "
+                f"'{self.namespace}'."
+            )
```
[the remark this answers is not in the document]
ehanson8: This block ensure all metrics are being sent to the same namespace and raises an error if not:

```
# Ensure all metrics resolve to the same namespace
if len(namespaces) > 1:
    raise ValueError(  # noqa: TRY301
        f"Cannot publish metrics with different namespaces in a single "
        f"request. Found: {namespaces}"
    )
```
This app has only 1 namespace so I think it would be better to wait for multiple namespaces to be a factor before deciding if/how we want to handle them. 

And not sure why this showed up as not resolved for you, it is resolved in the GH UI. Strange!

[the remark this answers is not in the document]
ehanson8: I see what happened now, I ran this review before I pushed my commit so it repeated the same comments as the previous review. I hid it but apparently it could still be found!

dsc/workflows/base/workflow.py
ehanson8: Great call, added!

jonavellecuerdo:
@ehanson8 Resubmitting this comment as it may have been missed due to formatting issues:

I previously saw that [this comment](https://github.com/MITLibraries/dspace-submission-composer/pull/226#discussion_r3320522858) was not resolved. After reading it and looking at the code closely, this is what I gathered:

1. [This `for` loop](https://github.com/MITLibraries/dspace-submission-composer/blob/metrics/dsc/utils/aws/metrics.py#L151) will iterate over a list `<=CLOUDWATCH_METRICS_LIMIT` (i.e., 1K) 
2. [For each metric](https://github.com/MITLibraries/dspace-submission-composer/blob/metrics/dsc/utils/aws/metrics.py#L153-L154), it retrieves `metric.namespace` or the default value from `self.namespace` and adds it to the `namespaces` list. This means namespaces are _per_ metric. 🤔 
3. As the Copilot comment above reads, [the call to `self._cloudwatch.put_metric_data`](https://github.com/MITLibraries/dspace-submission-composer/blob/metrics/dsc/utils/aws/metrics.py#L175-L179) only uses the _last_ value from the `namespaces` list (via `namespaces.pop()`).

Question:
* Are all values in `namespaces` expected to be the same? 
* What happens if one of the metrics in `metric_data` have a different `namespace`?

ehanson8:
> @ehanson8 Resubmitting this comment as it may have been missed due to formatting issues:
 
I did reply on Friday but it got buried and I can't easily link to it, so copying here:

This block ensure all metrics are being sent to the same namespace and raises an error if not:

```
# Ensure all metrics resolve to the same namespace
if len(namespaces) > 1:
    raise ValueError(  # noqa: TRY301
        f"Cannot publish metrics with different namespaces in a single "
        f"request. Found: {namespaces}"
    )
```

This app has only 1 namespace so I think it would be better to wait for multiple namespaces to be a factor before deciding if/how we want to handle them.

And not sure why this showed up as not resolved for you, it is resolved in the GH UI. Strange!

**Answer**: I see what happened now, I ran this review before I pushed my commit so it repeated the same comments as the previous review. I hid it but apparently it could still be found!

jonavellecuerdo (approved):

PR #235, opened 2026-05-06
In 1626 digitized theses transformer

jonavellecuerdo:

This PR introduces a metadata transformer for the digitized theses workflow. This also initializes the workflow directory for digitized theses. As [noted in IN-1626](https://mitlibraries.atlassian.net/browse/IN-1626?focusedCommentId=185871), it became clear after some exploration that the XSLT provided by Atmire could not be easily "plugged into" the DSC infrastructure, with a key finding that the XSLT likely ran in a Java-based environment. Creating the metadata transformer required translating the derivations in the XSLT to Python code. 

Key things to note: 

1. [Cross-walk for `dc.contributor.department` fields](https://github.com/MITLibraries/dspace-submission-composer/blob/IN-1626-digitized-theses-transformer/dsc/workflows/digitized_theses/config/departments_crosswalk.json): The "v1" workflow relied on a CSV file that mapped variations of different department values derived using the XSLT to _normalized_ department strings. For simplicity, I converted the original CSV file to a JSON file. 
   1. There was only one instance of a duplicate in the original crosswalk CSV file. Will ask stakeholders what mapping should be applied; I expect that the latter mapping is what gets used:
      ```csv
      "Joint Program in Oceanography (Massachusetts Institute of Technology, Dept. of Earth, Atmospheric, and Planetary Sciences; and the Woods Hole Oceanographic Institution)",EAPS||OceanEng
      "Joint Program in Oceanography (Massachusetts Institute of Technology, Dept. of Earth, Atmospheric, and Planetary Sciences; and the Woods Hole Oceanographic Institution)",OceanEng
      ```
   1. I tried passing the CSV to an LLM to convert exact matches to regex expressions, but I decided to move forward with the current approach of exact-matching for now. 
1. [Changes to `test/` directory](https://github.com/MITLibraries/dspace-submission-composer/tree/IN-1626-digitized-theses-transformer/tests/workflows/digitized_theses): I think the `test/` directory should follow the structure of the app folder. Following the structure will make it easier to navigate unit tests. My hope is to do the same for the unit tests of other workflows soon.
1. [Unit tests and coverage](https://github.com/MITLibraries/dspace-submission-composer/actions/runs/25446637551/job/74652439246): I added only a few unit tests for `DigitizedThesesTransformer`, focusing on the **key** methods in the transformer class (i.e., the methods shared/used by field methods). This decision results in a coverage of 70% because the individual field methods lack unit tests.
   1. I did not want to provide unit tests for each field method, given the number of fields. In `transmogrifier`, doing so resulted in very long [test modules ](https://github.com/MITLibraries/transmogrifier/blob/main/tests/sources/xml/test_marc.py), and I'm curious if there is another way we can think about testing transformers.
   2. Given that this is a first pass of the transformer and getting the transformer into the app will allow us to build the rest of the workflow, I'm requesting that the reduced coverage not be a blocker for this PR. 😅 

1. There will be a separate PR to derive the non-QDC field `mit.theses.degree` using another configuration file.

I've been in discussions with @ghukill on how we can run the transformer on all the original MARC XML records from Alma for all the items in the frozen DSpace 8 "MIT Theses" community, **but** for purposes of this PR, reviewers can:

1. Review the added unit tests.
2. Review side-by-side comparison for **five** theses that were ingested to DSpace via the current v1 workflow (provided by Sadie).
   1. The files suffixed "-derived" represent the metadata generated by the transformer.
   1. The files without the suffix represent the reformatted **target** metadata retrieved from DSpace (reformatted to `key: [list of values]`). Note the presence of `dc.description.provenance` in the metadata retrieved from DSpace; these were added to the record by Atmire when items were ingested via the v1 workflow which is why they are present in the transformer-derived metadata. 
     
      1. OCLC number **05588126**: 
[05588126-derived.json](https://github.com/user-attachments/files/27450015/05588126-derived.json) | 
[05588126.json](https://github.com/user-attachments/files/27450030/05588126.json)

      1. OCLC number **12854433**:  
[12854433-derived.json](https://github.com/user-attachments/files/27450063/12854433-derived.json) | 
[12854433.json](https://github.com/user-attachments/files/27450075/12854433.json)

      1. OCLC number **22215428**: 
[22215428-derived.json](https://github.com/user-attachments/files/27450080/22215428-derived.json) | 
[22215428.json](https://github.com/user-attachments/files/27450084/22215428.json)

      1. OCLC number **25395837**: 
[25395837-derived.json](https://github.com/user-attachments/files/27450090/25395837-derived.json) | 
[25395837.json](https://github.com/user-attachments/files/27450094/25395837.json)

      1. OCLC number **34619573**:  
[34619573-derived.json](https://github.com/user-attachments/files/27450097/34619573-derived.json) | 
[34619573.json](https://github.com/user-attachments/files/27450101/34619573.json)

Appreciate your patience on the development of this workflow!

YES - Updated dependencies for resolved CVEs; new transformer class relies on built-in `lxml` package

NO

### Code review


tests/workflows/digitized_theses/test_transformer.py
ghukill, line 1: Based on your #3 testing comments in the PR notes, I fully agree that testing _every_ method is not a good ROI.

Transmogrifier is aguably a bit different because a) not all sources use all fields, b) Transmog basically has **one** job and that's to transform.

This transformer here is very important, but it's a piece in a larger puzzle of preparing items for ingest.

dsc/workflows/digitized_theses/transformer.py
```diff
@@ -0,0 +1,1045 @@
...
+
+from lxml import etree
+
+from dsc.config import load_external_config
+
+NSMAP = {"marc": "http://www.loc.gov/MARC21/slim"}
+
+
+class DigitizedThesesTransformer:
+    """Transformer for Digitized Theses source metadata."""
```
ghukill: I think I'd propose moving some of the docstring information from the `transform()` method here.  I'd expect the following:

- class level docstring: why does this class exist?  what can it do?  what metadata format to what metadata format?
- `transform` docstring: what implementation details are relevant, that the docstring doesn't cover?  how does it technically achieve the purpose of the class, if helpful to notate?

I think a major class like this deserves a docstring that introduces a code reader to why it's here.

dsc/workflows/digitized_theses/transformer.py
```diff
@@ -0,0 +1,1045 @@
...
+        recs = root.xpath("//marc:record", namespaces=NSMAP)
+        if recs:
+            return recs[0]
+        return None
+
+    @staticmethod
+    def _xpath(
+        element: etree._Element,
+        expr: str,
+    ) -> list:
```
ghukill, line 135: Small, but having done lots of xpath work in python transformers, this is a nice touch.

dsc/workflows/digitized_theses/transformer.py
ghukill: As I read through the code, circling back here. 

I'm noticing -- and you mentioned in a discussion -- that this class is very `@staticmethod` oriented.  I'm appreciating that methods are generally expecting a `record: etree._Element` as input to operate over.  

This itself implies that this transformer is always expecting a **parsed** XML record as input.  This is the kind of detail that I think is helpful in the class docstring.  Because, this has implications for how this will get used in the workflow itself.

dsc/workflows/digitized_theses/transformer.py
```diff
@@ -0,0 +1,1045 @@
...
+                    parts.append(text)
+        return separator.join(parts).strip()
+
+    # =================================
+    #  DC element (qualified) methods
+    # =================================
+
+    @classmethod
+    def dc_title(cls, record: etree._Element) -> list[str]:
+        """MARC 245 $a $b  and  246 $a $b."""
```
ghukill: _I'm going to pick on this metadata field method, but applies to all._

I.... think I'm okay with the MARC tags that are targeted by the method as the only detail here?  For this method specifically, there is an implied relationship between the field method name `dc_title` and the MARC tags + indicators + subfields in the docstring `245...`.  

I'm inclined to thumbs up this approach with reservations.   I think a better docstring repeatable pattern might be something like:

```
Target dspace field: `dc.title`

<If ever helpful, more detail here...>

Tags used:
  - 245 $a $b
  - 246 $a $b
```

This is probably something an AI agent could slap together in a single request?   It's not required, but if we envision more transformers in the future, it could be a good pattern to set now.

dsc/workflows/digitized_theses/transformer.py
ghukill: Leaving the previous comment for posterity, as it kind of proves my point, but it was wrong!  I looked closer at `transform()` and see it's expecting `source_metadata: str | bytes` as input.  

I don't want to be overly prescriptive on the relationship of the class and `transform()` (main entrypoint) docstrings, but just a little help for building an intuition about what the class is and how it operates would be helpful.

dsc/workflows/digitized_theses/transformer.py
```diff
@@ -0,0 +1,1045 @@
...
+        for field in cls.fields:
+            field_method = getattr(cls, field)
+            formatted_field_name = field.replace("_", ".")
+
+            # check if 'source_metadata' is in signature
+            signature = inspect.signature(field_method)
+            if "record" in signature.parameters:
+                transformed_metadata[formatted_field_name] = field_method(record)
+            else:
+                transformed_metadata[formatted_field_name] = field_method()
```
ghukill: This is clever, but maybe the comment is wrong?  Looking at the field methods, isn't it looking for the presence of `record` (an `etree._Element` object)?

Possible small tweak to comment + flow:

```python
# if field method requires the record XML, pass it            
if "record" in inspect.signature(field_method).parameters:
    transformed_metadata[formatted_field_name] = field_method(record)
# else, run field method without input
else:
    transformed_metadata[formatted_field_name] = field_method()
```

ghukill (commented):
Overall, great work!  It's one of those PRs with lots of lines... but the majority of them are just repeating field method logic.

The real architectual add feels like the class + transform ergonomics.

I've left a couple of comments about updating docstrings, and potentially some inline comments in `transform()`.  But ultimately, few to none suggestions on the code path itself.  

Looking good.

dsc/workflows/digitized_theses/transformer.py
jonavellecuerdo: @ghukill Thanks for calling out! Initially the docstrings included references to subfields, but maintaining that for all metadata field methods became tricky--especially when the derivation used >3 subfield codes.

To clarify, are you okay if the docstring for `dc_title`, specifically, simply read:

```python
""""Marc 245 and 246."""
```

While I do like the comment template above, I am wondering if it is redundant (i.e., should the code speak for itself)? 🤔 😅

dsc/workflows/digitized_theses/transformer.py
jonavellecuerdo: Definitely keeping this in my back pocket for future XML parsing. :D

dsc/workflows/digitized_theses/transformer.py
ghukill: I think it's tricky.  I agree that the docstring shouldn't outline all logic.  It follows that the tags are also present in code... but that's kind of a handy level of reference, "Ah oh okay, this field is using 6xx fields... I get it."

But I think there is a decent argument to be made that _none_ of them need docstrings unless they do something unique / unusual / unexpected / complex.

Open to anything!  Just noting.  Heck, maybe as-is is already a good balance 😄.

dsc/workflows/digitized_theses/transformer.py
jonavellecuerdo: Good suggestion -- applied!

tests/workflows/digitized_theses/test_transformer.py
jonavellecuerdo: What are your thoughts regarding the reduced coverage statistic? 🤔 (Tagging @ehanson8 in thread as well!)

dsc/workflows/digitized_theses/transformer.py
jonavellecuerdo: Let me know what you think of the updated docstrings in https://github.com/MITLibraries/dspace-submission-composer/pull/235/commits/c1a1797c708e0bf0019118b34ad949cf46c9275c. 😄

tests/workflows/digitized_theses/test_transformer.py
ghukill: I have zero problem with it.  I think that low coverage number is accurate, and it's a good reminider if we ever revisit this.  If we were to artificially inflate it, that would just hide the fact we've made this decision.

ghukill (approved):
Approved!  Thanks for the docstring updates.  Seems like a great first pass, and we know there is opportunity for incremental changes as the workflow develops.

dsc/workflows/digitized_theses/transformer.py
jonavellecuerdo: Leaving as-is for now. Thank you! Open to updating it in the future if there are any changes made to derivations then.

dsc/workflows/digitized_theses/transformer.py
```diff
@@ -0,0 +1,1050 @@
...
+    # =================================
+    #  DC element (qualified) methods
+    # =================================
+
+    @classmethod
+    def dc_title(cls, record: etree._Element) -> list[str]:
+        """MARC 245 and 246."""
+        results: list[str] = []
+
+        for df in cls._datafields(record, "245"):
```
ehanson8, line 208: Not a blocker but I don't love the shift back to abbreviations (`ind1`, `sf`, `df`) because I think it makes this code less readable. They are fairly easy to figure out but `indicator_1` `subfield`, and `datafield` immediately remove all ambiguity

dsc/workflows/digitized_theses/transformer.py
```diff
@@ -0,0 +1,1050 @@
...
+        for df in cls._datafields(record, "700"):
+            value = cls._subfield_text(df, "abd")
+            if value:
+                results.append(value)
+
+        return results or None
+
+    @classmethod
+    def dc_contributor_department(cls, record: etree._Element) -> list[str] | None:
+        """MARC 710 or 502 (if 710a or 710b is not set)."""
```
ehanson8, line 261: You might add a reference to `cls._normalize_departments(results)` in the docstring

tests/workflows/digitized_theses/test_transformer.py
ehanson8: Agree with @ghukill, leave it obvious so we can address in the future if we choose to

ehanson8 (approved):
Approving but give some thought to the abbreviations, I think it's adding a minor but unnecessary mental load on anyone new to the codebase and I think there's plenty of space for the full names

dsc/workflows/digitized_theses/transformer.py
jonavellecuerdo: Noting and will update docstring in later PR.

dsc/workflows/digitized_theses/transformer.py
jonavellecuerdo: Will likely expand this in later PR. Thank you for bringing up. I may leave "ind1" and "ind2" as is to align with MARC XML terms, though!

PR #240, opened 2026-05-20
In 1626 digitized theses workflow create

jonavellecuerdo:

This PR implement batch preparation methods for digitized theses workflow. I went back-and-forth on whether I should send a PR with the entire workflow, but as I saw the changes that were piling up, I decided it would be best to request review in reasonably sized chunks. This also allows us to focus / reflect on the key differences for this workflow for each step of the DSC workflow.

Here are some key points to highlight about the `create` / `prepare_batch` methods for the digitized theses workflow:
1. The workflow has [a `minted_batch_id` attribute](https://github.com/MITLibraries/dspace-submission-composer/blob/IN-1626-digitized-theses-workflow-create/dsc/workflows/digitized_theses/workflow.py#L48-L54), which is comprised of the original batch ID and the current date timestamp.
1. [`DigitizedTheses.prepare_batch`](https://github.com/MITLibraries/dspace-submission-composer/blob/IN-1626-digitized-theses-workflow-create/dsc/workflows/digitized_theses/workflow.py#L95-L161) does not collect exceptions in the `errors` list like other workflows.
   1. **Why:** This allows DSC to proceed with writing all item submissions to the DynamoDB table, with any exceptions captured via the `ItemSubmission.status_details` field.
      1. When DSC sends the report, the resulting CSV file from DynamoDB will include all intended item submissions in the batch rather than creating a separate `errors.csv`.
       1. No `BatchCreationFailedError` is raised. **TBD when this error should be raised (if ever) for the workflow.**
1. After the `create` step, the minted batch ID is what gets passed in the latter calls to `submit` and `finalize`.
   1. For this reason, I left `DigitizedTheses.batch_path` as is and explicitly used `DigitizedTheses.minted_batch_id` when writing files to S3.
1. Why do we save the Alma metadata in XML files to S3?
   1. When we call `submit`, we can avoid repeated calls to Alma SRU and read the metadata from S3 instead. The saved XML files are also simplified from the original Alma SRU response (the saved file sets the `<record>` as the root of the XML file -- will no longer require [these steps](https://github.com/MITLibraries/dspace-submission-composer/blob/bc86c7713ffa7d014bf9a32034d002d0fb07db22/dsc/workflows/digitized_theses/transformer.py#L144-L146)).
      
While the test suite includes [several new unit tests for the simpler functions in `dsc.workflows.DigitizedTheses`](https://github.com/MITLibraries/dspace-submission-composer/blob/IN-1626-digitized-theses-workflow-create/tests/workflows/digitized_theses/test_workflow.py), I'm requesting that we move forward with the current set of unit tests **given the additional local testing with MinIO shared below**.

**Note:** In general, DSC has been a complex app to test. Recent changes have relied on actual runs with the DSO Step Function, but this is not possible until we have a test instance of DSpace 8 available. Revisiting the test suite for clean up / improve maintainability is definitely on track for future work! Thanking reviewers in advance for understanding.

### Local testing of `dsc.workflows.DigitizedTheses.prepare_batch`

**About the test data:**
These five files were provided by Sadie (i.e., items that were imported into DSpace via v1 workflow):

* **12854433**: https://hdl.handle.net/1721.1/157046	
* **25395837**: https://hdl.handle.net/1721.1/148716
* **34619573**: https://hdl.handle.net/1721.1/154113
* **22215428**: https://hdl.handle.net/1721.1/158471	
* **05588126**: https://hdl.handle.net/1721.1/157651

An additional file for a student-submitted electronic theses (i.e., should be skipped by workflow) is used for testing:
* **932127295**: http://hdl.handle.net/1721.1/100593

---

1. Created a batch (`batch-0`) in the digitized theses workspace S3 bucket:
image

2. After running `dsc.workflows.DigitizedTheses.prepare_batch` in IPython:

- Contents are synced to a minted batch folder **five item submissions** moved to a `replacement-theses/` subfolder and **one item submission** (a student-submitted electronic thesis) remains at unmoved.
image

YES 

NO

### Code review

### DIFF
--- a/dsc/config.py
+++ b/dsc/config.py
@@ -15,6 +15,10 @@ class Config:
         "S3_BUCKET_SUBMISSION_ASSETS",
         "SOURCE_EMAIL",
         "SQS_QUEUE_DSS_INPUT",
+        # workflow-specific
+        "DSPACE_CREDENTIALS",
+        "METADATA_API_URL",
+        "S3_BUCKET_DIGITIZED_THESES",
     ]
 
     OPTIONAL_ENV_VARS: Iterable[str] = [
@@ -76,6 +80,28 @@ class Config:
         if _excluded_loggers := os.getenv("WARNING_ONLY_LOGGERS"):
             return _excluded_loggers.split(",")
         return []
+
+    # Workflow-specific env vars
+    @property
+    def dspace_credentials(self) -> str:
+        value = os.getenv("DSPACE_CREDENTIALS")
+        if not value:
+            raise OSError("Env var 'DSPACE_CREDENTIALS' must be defined")
+        return value
+
+    @property
+    def metadata_api_url(self) -> str:
+        value = os.getenv("METADATA_API_URL")
+        if not value:
+            raise OSError("Env var 'METADATA_API_URL' must be defined")
+        return value
+
+    @property
+    def s3_bucket_digitized_theses(self) -> str:
+        value = os.getenv("S3_BUCKET_DIGITIZED_THESES")
+        if not value:
+            raise OSError("Env var 'S3_BUCKET_DIGITIZED_THESES' must be defined")
+        return value
 
     def check_required_env_vars(self) -> None:
         """Method to raise exception if required env vars not set."""
--- a/dsc/config.py
+++ b/dsc/config.py
@@ -83,11 +83,13 @@ class Config:
 
     # Workflow-specific env vars
     @property
-    def dspace_credentials(self) -> str:
+    def dspace_credentials(self) -> dict:
         value = os.getenv("DSPACE_CREDENTIALS")
         if not value:
             raise OSError("Env var 'DSPACE_CREDENTIALS' must be defined")
-        return value
+        credentials = json.loads(value)
+
+        return {"IR-8": credentials["ir-8"], "DDC-8": credentials["ddc-8"]}
 
     @property
     def metadata_api_url(self) -> str:
--- a/dsc/db/models.py
+++ b/dsc/db/models.py
@@ -21,6 +21,9 @@ ITEM_SUBMISSION_LOG_STR = (
 
 
 class ItemSubmissionStatus(StrEnum):
+    CREATE_SUCCESS = "create_success"
+    CREATE_FAILED = "create_failed"
+    CREATE_SKIPPED = "create_skipped"
     BATCH_CREATED = "batch_created"
     SUBMIT_SUCCESS = "submit_success"
     SUBMIT_FAILED = "submit_failed"
--- a/dsc/workflows/digitized_theses/workflow.py
+++ b/dsc/workflows/digitized_theses/workflow.py
@@ -1,0 +1,459 @@
+# ruff: noqa: FIX002, TD002, TD003
+import glob
+import logging
+import os
+import re
+import shutil
+import tempfile
+from collections.abc import Iterator
+from enum import StrEnum
+from pathlib import Path
+from typing import Any
+
+import requests
+from dspace_rest_client.client import DSpaceClient
+from dspace_rest_client.models import Item as DSpaceItem
+from lxml import etree
+
+from dsc import exceptions
+from dsc.config import Config
+from dsc.db.models import ItemSubmissionStatus
+from dsc.item_submission import ItemSubmission
+from dsc.utils.aws.s3 import S3Client, run_aws_cli_sync
+from dsc.workflows.base import Workflow
+from dsc.workflows.digitized_theses import NSMAP, DigitizedThesesTransformer
+
+CONFIG = Config()
+logger = logging.getLogger(__name__)
+
+
+class MITThesesCommunityUUID(StrEnum):
+    DEV = ""  # TODO: Update to uuid for 'MIT Theses' community in test
+    PROD = "6fc02cc2-0d14-4023-8a6f-d9900d0c4302"
+
+
+class DigitizedTheses(Workflow):
+    """Workflow for digitized theses from the Imaging Lab.
+
+    This workflow is unique to other workflows in that it relies on an
+    additional S3 bucket that serves as a "workspace" for digitized theses.
+    Users create a batch folder in this workspace S3 bucket and upload
+    theses (PDF files) named with the OCLC number: <oclc-number>-MIT.pdf.
+
+    When this workflow creates a batch *without* syncing, it will first
+    copy the contents of the original batch folder in the workspace S3 bucket
+    to a dated batch folder in a temporary directory on local disk.
+    The batch folder will organize item submission assets into subfolders
+    based on whether it is a new or replacement thesis or whether it should
+    be skipped. Once the contents of the local batch are ready, it is
+    synced to a similarly named batch folder in the DSC S3 bucket and
+    ItemSubmission's are recorded in DynamoDB.
+
+    When this workflow creates a batch *with* syncing, this means an organized
+    batch folder already exists in the DSC S3 bucket in Prod, which is the
+    result of syncing data from Stage. This method will instead walk
+    the contents of the synced batch folder to create ItemSubmission's
+    that are recorded in DynamoDB.
+
+    TODO: Add description for `submit` and `finalize` methods.
+    """
+
+    workflow_name: str = "digitized-theses"
+    metadata_transformer = DigitizedThesesTransformer
+
+    def __init__(self, batch_id: str):
+        self._dspace_client = None
+        self.community_uuid = (
+            MITThesesCommunityUUID.PROD
+            if CONFIG.workspace == "prod"
+            else MITThesesCommunityUUID.DEV
+        )
+        super().__init__(batch_id)
+
+    @property
+    def metadata_mapping_path(self) -> str:
+        raise NotImplementedError
+
+    @property
+    def dspace_client(self) -> DSpaceClient:
+        if not self._dspace_client:
+            logger.debug(
+                f"Creating DSpace client for destination {self.submission_system}"
+            )
+
+            try:
+                credentials = CONFIG.dspace_credentials[self.submission_system]
+            except KeyError as exception:
+                raise exceptions.DSpaceClientCredentialsNotFoundError(
+                    f"No credentials for {self.submission_system}"
+                ) from exception
+
+            client = DSpaceClient(
+                api_endpoint=credentials["url"],
+                username=credentials["user"],
+                password=credentials["password"],
+                fake_user_agent=True,
+            )
+            authenticated = client.authenticate()
+            if not authenticated:
+                raise exceptions.DSpaceClientAuthenticationError(
+                    credentials["url"], credentials["user"]
+                )
+            self._dspace_client = client
+            logger.info(
+                f"Successfully authenticated to {credentials['url']} "
+                f"as {credentials['url']}"
+            )
+
+        return self._dspace_client
+
+    def get_batch_bitstream_uris(self) -> list[str]:
+        raise NotImplementedError
+
+    def item_metadata_iter(self) -> Iterator[dict[str, Any]]:
+        raise NotImplementedError
+
+    def prepare_batch(self, *, synced: bool = False) -> tuple[list, ...]:
+        """Prepare a batch folder in the DSC S3 bucket.
+
+        Unlike other workflows, successes, failures, and skips are recorded in the
+        item submissions table in DynamoDB, as a means to collate all results for a batch
+        in the batch creation report. This means, the method will always return an empty
+        'errors' list.
+        """
+        item_submissions = []
+        errors: list[tuple] = []  # set but not used
+
+        if synced:
+            logger.info(f"Batch folder already created in WORKSPACE={CONFIG.workspace}")
+            item_submissions = self._get_item_submissions_from_synced_batch()
+        else:
+            item_submissions = self._create_batch_in_s3()
+
+        return item_submissions, errors
+
+    def _get_item_submissions_from_synced_batch(self) -> list[ItemSubmission]:
+        """Create ItemSubmission's from a synced batch folder in the DSC S3 bucket.
+
+        This method loops through the uris for objects in the synced batch folder.
+        From each uri, the method extracts the theses subfolder and the item identifier
+        for an item submission. The theses subfolder determines the ItemSubmission.status
+        to set in DynamoDB:
+            - Item submissions in replacement-theses or new-theses -> CREATE_SUCCESS
+            - Item submissions in skipped-theses -> CREATE_SKIPPED
+
+        As the method loops through the uris, it checks against a list of seen
+        item identifiers to avoid duplicates.
+        """
+        item_submissions = []
+        s3_client = S3Client()
+
+        # pattern to extract meta about item submission from uri
+        pattern = r"/(?:[^/]+/)*?([a-zA-Z0-9-]*-theses)/(\d+)(?:/.*)?$"
+
+        seen_item_identifiers = []
+        for uri in s3_client.files_iter(bucket=self.s3_bucket, prefix=self.batch_path):
+            match = re.search(pattern, uri)
+            if not match:
+                logger.warning(f"Cannot create item submission for uri: {uri}")
+                continue
+
+            # retrieve theses subfolder and item identifier from uri
+            theses_subfolder, item_identifier = match.groups()
+            if item_identifier in seen_item_identifiers:
+                logger.debug(
+                    f"Already created an item submission for item_identifier={item_identifier}"  # noqa: E501
+                )
+                continue
+
+            seen_item_identifiers.append(item_identifier)
+            item_submissions.append(
+                ItemSubmission(
+                    batch_id=self.batch_id,
+                    item_identifier=item_identifier,
+                    workflow_name=self.workflow_name,
+                    status=(
+                        ItemSubmissionStatus.CREATE_SUCCESS
+                        if theses_subfolder in ["replacement-theses", "new-theses"]
+                        else ItemSubmissionStatus.CREATE_SKIPPED
+                    ),
+                )
+            )
+
+        return item_submissions
+
+    def _create_batch_in_s3(self) -> list[ItemSubmission]:
+        """Create a batch of item submissions in the DSC S3 bucket.
+
+        This method begins with syncing the contents of the original batch folder
+        in the digitized-theses-workspace S3 bucket to a dated batch folder in
+        a temporary directory on local disk. From there, it performs the following
+        steps for each thesis:
+
+        1. Download the MARC XML record from Alma and save to an XML file
+        2. Check if an item in DSpace already exists:
+            - If it exists and it is a student-submitted electronic thesis, skip
+            - If it exists and is not, mark as 'Replacement thesis'
+            - If it does not exist, mark as 'New thesis'
+        3. Organize the contents of the dated batch folder by thesis type
+        4. Sync the dated batch folder to the workflow folder in the DSC S3 bucket
+        5. Clean up the temporary directory
+        """
+        item_submissions = []
+
+        # update self.batch_id to include current date timestamp
+        original_batch_id = self.batch_id
+        self.batch_id = self._update_batch_id(original_batch_id)
+
+        # create temp directory
+        tmp_dir = tempfile.TemporaryDirectory()
+        logger.info(f"Created temporary directory: {tmp_dir.name}")
+
+        # sync batch folder from digitized-theses-workspace S3 bucket
+        # to dated batch folder in temp directory
+        tmp_batch_path = f"{tmp_dir.name}/{self.batch_id}"
+        run_aws_cli_sync(
+            source=f"s3://{CONFIG.s3_bucket_digitized_theses}/{original_batch_id}",
+            destination=tmp_batch_path,
+        )
+
+        for file in os.listdir(tmp_batch_path):
+            item_submission = ItemSubmission(
+                batch_id=self.batch_id,
+                item_identifier=self.parse_item_identifier(file),
+                workflow_name=self.workflow_name,
+            )
+
+            # get MARC XML metadata from Alma
+            try:
+                self._download_metadata_from_alma(
+                    item_submission, batch_location=tmp_batch_path
+                )
+            except (
+                requests.exceptions.HTTPError,
+                exceptions.ItemMetadataNotFoundError,
+            ) as exception:
+                item_submission.status = "create_failed"
+                item_submission.status_details = str(exception)
+                item_submissions.append(item_submission)
+                continue
+
+            # check if item with the OCLC number exists
+            try:
+                dspace_item = self._get_item_from_dspace(item_submission.item_identifier)
+            except exceptions.DSpaceClientSearchError as exception:
+                item_submission.status = "create_skipped"
+                item_submission.status_details = str(exception)
+                item_submissions.append(item_submission)
+                continue
+
+            # check if item submission is a 'Replacement thesis'
+            if dspace_item and not self._is_replacement_thesis(dspace_item):
+                item_submission.status = "create_skipped"
+                item_submission.status_details = "Cannot replace the electronic version submitted by the student author."  # noqa: E501
+                item_submissions.append(item_submission)
+                continue
+
+            if dspace_item and self._is_replacement_thesis(dspace_item):
+                item_submission.status = "create_success"
+                item_submission.status_details = "Replacement thesis"
+                item_submissions.append(item_submission)
+            else:
+                item_submission.status = "create_success"
+                item_submission.status_details = "New thesis"
+                item_submissions.append(item_submission)
+
+        self._move_batch_files_to_theses_subfolders(
+            item_submissions, batch_location=tmp_batch_path
+        )
+
+        # sync batch folder from digitized-theses-workspace S3 bucket
+        # to dated batch folder in temporary directory
+        run_aws_cli_sync(
+            source=tmp_batch_path,
+            destination=f"s3://{CONFIG.s3_bucket_submission_assets}/{self.batch_path}",
+        )
+
+        # clean up temp directory
+        tmp_dir.cleanup()
+
+        return item_submissions
+
+    def _update_batch_id(self, batch_id: str) -> str:
+        """Create a new batch ID with a date timestamp.
+
+        This method is only used when creating a batch *without syncing*.
+        The updated batch ID is used to distinguish different runs of
+        Workflow.create_batch, which can be run as many times as needed
+        until a batch is ready for submission.
+        """
+        return f"{batch_id}-{self.run_date.strftime('%Y%m%dT%H%M%SZ')}"
+
+    def _download_metadata_from_alma(
+        self, item_submission: ItemSubmission, batch_location: str
+    ) -> bytes:
+        """Download MARC XML metadata for an item submission from Alma.
+
+        This method writes an XML file with MARC metadata to the
+        batch folder in S3 named: '<oclc-number>-MIT.xml'
+
+        For more information, see:
+        https://developers.exlibrisgroup.com/alma/integrations/sru/.
+        """
+        logger.debug(
+            f"Retrieving metadata from Alma for an item with alma.oclc_control_number_035_a={item_submission.item_identifier}"  # noqa: E501
+        )
+
+        query_url = f"https://{CONFIG.metadata_api_url}"
+        response = requests.get(
+            query_url,
+            params={
+                "operation": "searchRetrieve",
+                "recordSchema": "marcxml",
+                "query": (
+                    f"alma.oclc_control_number_035_a={item_submission.item_identifier}"
+                ),
+            },
+            timeout=180,
+        )
+
+        try:
+            response.raise_for_status()
+        except requests.exceptions.HTTPError as exception:
+            # TODO: Custom error includes HTTPError message
+            raise exceptions.ItemMetadataNotFoundError from exception
+
+        # get nested `<marc:record>` element from SRU response
+        record_element = self._parse_record_from_sru_response(response.content)
+
+        # write an XML file to batch folder
+        with open(
+            Path(batch_location) / f"{item_submission.item_identifier}.xml", "wb"
+        ) as file:
+            tree = etree.ElementTree(record_element)
+            tree.write(file, xml_declaration=True, encoding="UTF-8")
+
+        return response.content
+
+    def _get_item_from_dspace(self, item_identifier: str) -> DSpaceItem:
+        """Get Item object from DSpace given an item identifier.
+
+        This method expects only a single Item object for a given item identifier
+        (if any). If DSpaceClient.search_objects returns None or a list of more than
+        one Item objects, this method raises exceptions.DSpaceClientSearchError.
+        """
+        logger.debug(
+            f"Searching DSpace for an item with dc.identifier.oclc={item_identifier}"
+        )
+
+        dspace_objects = self.dspace_client.search_objects(
+            query=f"dc.identifier.oclc:{item_identifier}",
+            scope=self.community_uuid,
+            dso_type="item",
+        )
+
+        if dspace_objects is None:
+            raise exceptions.DSpaceClientSearchError(
+                f"Failed search for item with dc.identifier.oclc={item_identifier} "
+                f"in community {"<community>"}"
+            )
+        if len(dspace_objects) > 1:
+            raise exceptions.DSpaceClientSearchError(
+                f"Expecting one item with dc.identifier.oclc={item_identifier} "
+                f"in community {"<community>"}; found {len(dspace_objects)} items"
+            )
+        if len(dspace_objects) == 0:
+            return None
+
+        return dspace_objects[0]
+
+    def _move_batch_files_to_theses_subfolders(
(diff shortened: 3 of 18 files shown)


dsc/workflows/digitized_theses/workflow.py
```diff
@@ -0,0 +1,328 @@
...
+            authenticated = client.authenticate()
+            if not authenticated:
+                raise exceptions.DSpaceClientAuthenticationError(
+                    credentials["url"], credentials["user"]
+                )
+            self._dspace_client = client
+            logger.info(
+                f"Successfully authenticated to {credentials['url']} "
+                f"as {credentials['url']}"
+            )
```
jonavellecuerdo, line 106: Pulled from DSS!

ghukill:
Before I jump into a review @jonavellecuerdo, had a question about your second screenshot.

For PDF `932127295`, am I understand correctly that:
- the batch folder is created in DSC bucket
- the original PDF is moved over to root of batch folder
- the Alma MARC is retrieved and saved as an XML document
- but because not replacement or new, it remains in the root of the batch folder

Is that right?  I suppose my naive expectation would be that bad/error/skipped records would get moved into a folder like "skipped".  This might be a decent name, is it captures true skips, unhandled errors, etc.  

It's a bit minor, but it feels kind of messy to have the "bad" PDF + XMLs at the root of the batch folder.  Open to pushback or clarification here, just a thought as I prepare to review tomorrow!

dsc/config.py
```diff
@@ -77,6 +81,28 @@ def warning_only_loggers(self) -> list:
...
+        if not value:
+            raise OSError("Env var 'DSPACE_CREDENTIALS' must be defined")
+        return value
+
+    @property
+    def metadata_api_url(self) -> str:
+        value = os.getenv("METADATA_API_URL")
+        if not value:
+            raise OSError("Env var 'METADATA_API_URL' must be defined")
+        return value
```
ehanson8, line 99: So this would be used by both DT and Wiley for example?

dsc/config.py
```diff
@@ -77,6 +81,28 @@ def warning_only_loggers(self) -> list:
...
+
+    @property
+    def s3_bucket_digitized_theses(self) -> str:
+        value = os.getenv("S3_BUCKET_DIGITIZED_THESES")
+        if not value:
+            raise OSError("Env var 'S3_BUCKET_DIGITIZED_THESES' must be defined")
+        return value
```
ehanson8, line 106: Similarly to above (if I'm understanding it right), could this also be more generic, e.g. `working_directory`?

dsc/config.py
```diff
@@ -77,6 +81,28 @@ def warning_only_loggers(self) -> list:
...
+
     def check_required_env_vars(self) -> None:
         """Method to raise exception if required env vars not set."""
         missing_vars = [var for var in self.REQUIRED_ENV_VARS if not os.getenv(var)]
```
ehanson8, line 110: I'm still a bit unclear how DSC would manage workflow-specific env vars, I get the need to hit different buckets for different AWS envs for `working_directory` but it is odd to have a workflow-specific env vars at the `Config` level. That said, if the plan is to create a ticket and address later, that's OK with me. Just want to avoid a proliferation of workflow-specific env var in the future

For `metadata_api_url` however, a workflow class attribute feels like a better fit in most, if not all, cases. Even with Alma, I don't think we would read from Alma sandbox even on `Dev1`, we would read from Alma prod.

dsc/db/models.py
```diff
@@ -21,6 +21,9 @@
 
 
 class ItemSubmissionStatus(StrEnum):
+    CREATE_SUCCESS = "create_success"
+    CREATE_FAILED = "create_failed"
+    CREATE_SKIPPED = "create_skipped"
     BATCH_CREATED = "batch_created"
```
ehanson8, line 27: `BATCH_CREATED` feels duplicative now, do we still need it? And if we are changing enum values for DynamoDB, we may need to run an update pass to ensure consistency

dsc/workflows/digitized_theses/workflow.py
```diff
@@ -0,0 +1,328 @@
...
+from dsc.workflows.base import Workflow
+from dsc.workflows.digitized_theses import NSMAP, DigitizedThesesTransformer
+
+CONFIG = Config()
+logger = logging.getLogger(__name__)
+
+
+class MITThesesCommunityUUID(StrEnum):
+    dev = ""  # TODO: Update to uuid for 'MIT Theses' community in test
+    prod = "6fc02cc2-0d14-4023-8a6f-d9900d0c4302"
```
ehanson8: As enums, should these be `DEV` and `PROD`?

dsc/workflows/digitized_theses/workflow.py
```diff
@@ -0,0 +1,328 @@
...
+                continue
+
+            # check if item submission is a 'Replacement thesis'
+            if dspace_item and not self._is_replacement_thesis(dspace_item):
+                item_submission.status = "create_skipped"
+                item_submission.status_details = "Cannot replace the electronic version submitted by the student author."  # noqa: E501
+                item_submissions.append(item_submission)
+                continue
+
+            if dspace_item and self._is_replacement_thesis(dspace_item):
```
ehanson8, line 257: Should the 2nd conditional be `elif`? You could also consider putting the happy path conditional first but I don't think we have a settled best practice on that

dsc/workflows/digitized_theses/workflow.py
```diff
@@ -0,0 +1,328 @@
...
+                item_submission.status_details = "Replacement thesis"
+                item_submissions.append(item_submission)
+            else:
+                item_submission.status = "create_status"
+                item_submission.status_details = "New thesis"
+                item_submissions.append(item_submission)
+
+        self._move_batch_files_to_theses_subfolders(item_submissions)
+
+        return item_submissions, errors
```
ehanson8: This method is really well-laid out and easy-to-read, great work!

dsc/workflows/digitized_theses/workflow.py
```diff
@@ -0,0 +1,328 @@
...
+    def _get_item_from_dspace(self, item_submission: ItemSubmission) -> DSpaceItem:
+        logger.debug(
+            f"Searching DSpace for an item with dc.identifier.oclc={item_submission.item_identifier}"  # noqa: E501
+        )
+
+        dspace_objects = self.dspace_client.search_objects(
+            query=f"dc.identifier.oclc:{item_submission.item_identifier}",
+            scope=self.community_uuid,
+            dso_type="item",
+        )
```
ehanson8, line 353: Awesome, I remember when you were asking about this!

dsc/workflows/digitized_theses/workflow.py
```diff
@@ -0,0 +1,328 @@
...
+        """Determine if a DSpace item is a replacement thesis.
+
+        If an item already exists in DSpace, the thesis (bitstream) should
+        only be replaced *if* it is not a student-submitted electronic PDF
+        file, which is checked via the presence of a target string in the
+        dc.description field.
+
+        This method returns True when the target string does *not* appear in
+        any dc.description value (if any), indicating that
+        the replacement is valid.
```
ehanson8: Great context for an odd check we have to do! 😆

dsc/workflows/digitized_theses/workflow.py
```diff
@@ -0,0 +1,328 @@
...
+        if int(number_of_records.text) > 1:
+            # TODO: Custom error includes message
+            # "Unexpected response from Alma SRU, multiple records with OCLC"
+            raise exceptions.ItemMetadataNotFoundError
+
+        return root.xpath("//marc:record", namespaces=NSMAP)[0]
+
+    @staticmethod
+    def parse_item_identifier(filename: str) -> str:
+        return filename.rsplit("/", maxsplit=1)[-1].removesuffix("-MIT.pdf")
```
ehanson8, line 459: This whole module is very clear and a pleasure to read!

tests/workflows/digitized_theses/test_workflow.py
```diff
@@ -0,0 +1,199 @@
...
+    mock_dspace_client.authenticate.assert_called_once()
+    patched_dspace_client.assert_called_once_with(
+        api_endpoint="mock://mit-dspace.test.4science.cloud/server/api",
+        username="user@test.com",
+        password="topsecret",  # noqa: S106
+        fake_user_agent=True,
+    )
+
+
+def test_workflow_dspace_client_raise_authenticationerror(
```
ehanson8: `test_workflow_dspace_client_raise_authentication_error`?

ehanson8 (commented):
Looking good! Some comments and questions

dsc/workflows/digitized_theses/workflow.py
```diff
@@ -0,0 +1,328 @@
...
+        (a PDF) represents a digitized thesis and is assigned the filename
+        '<oclc-number>-MIT.pdf', which yield item identifiers formatted as
+        '<oclc-number>-MIT'.
+
+        Unlike other workflows, successes, failures, and skips are recorded in the
+        item submissions table in DynamoDB, as a means to collate all results for a batch
+        in the batch creation report. This means, the method will always return an empty
+        'errors' list.
+        """
+        item_submissions = []
```
ghukill, line 124: As discussed in an off-PR huddle, I might propose skipping some of the "minted" batch id properties + logic,

```python
@property
    def minted_batch_id(self) -> str:
```

And consider just **modifying** the batch id **only** for the `prepare_batch()` work, e.g.:

```python
self.batch_id = self.set_batch_id(batch_id)
```

where `.set_batch_id()` might have docstrings to explain that the "raw" batch id we get at the very beginning is not sufficient, so we have to modify or "set" it for all future work.  Maybe noting that for future work like submit + finalize, the batch id is already "set" and comes from Dynamo.

dsc/workflows/digitized_theses/workflow.py
```diff
@@ -0,0 +1,328 @@
...
+        errors: list[tuple] = []  # set but not used
+
+        # sync batch folder from DT S3 bucket to minted batch folder in DSC S3 bucket
+        run_aws_cli_sync(
```
ghukill, line 215: Also touched on in an off-PR discussion, I think there is some conflict with the word "sync" here, that may or may not be worth addressing.  

In the method signature `synced: bool = False` is referring to whether or not the batch folder was synced from stage --> prod, correct?  and that we're not heavily using this right now?

Either way, it seems as though this has some tension with the use of "sync" in the utility function `run_aws_cli_sync(...)` which while technically "syncing" data in S3, it's more of a naive copy from point A to point B.  My understanding is that it's not an "opinionated DSC sync" of a fully formed batch folder from stage --> prod.

I agree that "sync" is still accurate in that function name, given its literally running `aws cli sync...`.  But we may just need some comments + docstrings to help readers know super clearly that this particular workflow kind of has a **third** bucket in play, the digitized theses bucket that Jenn (or others) upload PDFs to.

I'm going to avoid needling the whole "do things in stage, then do again in prod" logic which I can appreciate but don't love here.  Assuming that remains and we like it, I still think there is room here to disambiguate these two things:

1. This `DigitizedTheses` workflow **copying** data from a workspace bucket specific to only this workflow, to the stage bucket
2. Syncing a fully formed batch folder from stage to prod

dsc/workflows/digitized_theses/workflow.py
```diff
@@ -0,0 +1,328 @@
...
+logger = logging.getLogger(__name__)
+
+
+class MITThesesCommunityUUID(StrEnum):
+    dev = ""  # TODO: Update to uuid for 'MIT Theses' community in test
+    prod = "6fc02cc2-0d14-4023-8a6f-d9900d0c4302"
+
+
+class DigitizedTheses(Workflow):
+    """Workflow for digitized theses from the Imaging Lab."""
```
ghukill: I would like to propose that we expand this docstrings pretty dramatically.  To me, this workflow class docstring could be the primary place where this workflow is explained.  Could touch on things like:

- modifying batch ids 
- special behavior of copying data from an specific S3 bucket (Jenn's digitized theses PDF bucket)
- metadata mechanics like getting Alma MARC, converting to DSpace DC, etc.
- identifying things which are unusual or unexpected

To me, the goal would be after a close read of this docstring you have footholds of understanding for the code as you read it.

This docstring doesn't need to be exhaustive of if/else logic, and should probably stay away from naming specific methods and functions and S3 locations.  It should really just explain at a high level how this workflow, works!  Please do pushback if we have another place for this information I'm not aware of.

dsc/workflows/digitized_theses/workflow.py
```diff
@@ -0,0 +1,328 @@
...
+        errors: list[tuple] = []  # set but not used
+
+        # sync batch folder from DT S3 bucket to minted batch folder in DSC S3 bucket
+        run_aws_cli_sync(
+            source=f"s3://{CONFIG.s3_bucket_digitized_theses}/{self.batch_id}",
+            destination=f"s3://{CONFIG.s3_bucket_submission_assets}/{self.workflow_name}/{self.minted_batch_id}",
+        )
+
+        s3_client = S3Client()
+        for file in s3_client.files_iter(
```
ghukill: Image

A lot of my comments are kind of around this file copying + syncing + batch folder creation, so this comment is actually kind of touching three things:

1. comment about possible "setting" of `self.batch_id`, which means `self.batch_id` could be used more
2. the use of `run_aws_cli_sync(...)`
3. this call to `s3_client.files_iter(...)`

Could there be sub-methods for some of this work?  So this `prepare_batch()` might look more like?

```python
self.batch_id = self.update_batch_id(batch_id)

self._copy_input_files_to_dsc_batch_folder(...)

input_files = self._get_input_files_iter(...)

item_submissions = []
errors: list[tuple] = []  # set but not used
for file in input_files:
    ...
```

Just throwing it out there, there might be some opportunity for small, targeted methods here.  So when encountering this `prepare_batch()` method, you can follow along at a high level.

The screenshot is meant to communicate thow _ideally_ there wouldn't be as much explicit S3 bucket + path construction at this high level.

dsc/config.py
jonavellecuerdo: Yes. 🤔 I think this is also worth asking @cabutlermit  about when we talk about how to update the `DSPACE_CREDENTIALS` environment variable (i.e., multiple credentials/info intended for a single environment variable).

dsc/config.py
jonavellecuerdo: I think it could be! Requesting to leave it for now until we get into the WIley implementation. What this is revealing is a need to reevaluate how to dynamically set the appropriate configs based on the workflow `METADATA_API_URL` and `S3_BUCKET_WORKING_DIRECTORY`. I think the discussion for `DSPACE_CREDENTIALS` will help us figure out how we should be storing these configs in AWS in the first place and from there we can figure out how the app should access them. 

Does that sound like a reasonable next step -- proceed as-is, include in discussion with @cabutlermit ? 🤔

dsc/config.py
ehanson8: Your comment is making me realize we could do a similar JSON blob structure for other env vars that change for various workflows, but yes, consulting with him would be great. This addresses my concerns a few comments down

dsc/config.py
jonavellecuerdo: Created this ticket for now! This has been on my mind as well but have been focused on getting something up and running first, figuring we can tackle infra-related changes after discussions with @cabutlermit !

dsc/config.py
ehanson8: Yes, I'm fine with that!

dsc/db/models.py
jonavellecuerdo: I left `ItemSubmissionStatus.BATCH_CREATED` to prevent breaking workflows that rely on it, but I did create a ticket to align all workflows to use the new statuses: [IN-1753](https://mitlibraries.atlassian.net/browse/IN-1753). I don't believe adding a new status changes anything in terms of writing to DynamoDB. As long as it meets [the type](https://github.com/MITLibraries/dspace-submission-composer/blob/IN-1626-digitized-theses-workflow-create/dsc/db/models.py#L100), it should write to DynamoDB just fine.

dsc/workflows/digitized_theses/workflow.py
jonavellecuerdo: Hmm, this gets used [during `DigitizedTheses.__init__` and retrieves the correct community UUID based on `Config.workspace`](https://github.com/MITLibraries/dspace-submission-composer/blob/IN-1626-digitized-theses-workflow-create/dsc/workflows/digitized_theses/workflow.py#L37-L41), which is based on the `WORKSPACE` env var, which is usually lowercase...

However reading docs does indicate these are "strongly recommended to be capitalized". I will update this!

dsc/workflows/digitized_theses/workflow.py
jonavellecuerdo: Ah! Ruff actually told me to get rid of the `elif` because of the `continue` at line 154. I also thought about putting the happy path conditional first, but I opted this sequence s.t. `continue` statements hit first (i.e., skip unneeded steps/checks early!

dsc/workflows/digitized_theses/workflow.py
jonavellecuerdo: Yay past us!

jonavellecuerdo: Definitely picking up patterns from our collective learnings. :)

dsc/workflows/digitized_theses/workflow.py
jonavellecuerdo: electronic pdf != digitized pdf 👁️ 

As I read this again, I will reword to:


```text
If an item is already present in DSpace, its thesis file (bitstream) 
should be replaced only when it is not a student-submitted electronic PDF. 
This determination is made by checking for a specific target string in the dc.description field.

This method returns True if the target string is absent 
from all dc.description entries (if any exist), indicating that it is 
safe to proceed with the replacement.
```

tests/workflows/digitized_theses/test_workflow.py
jonavellecuerdo: I always wonder whether it'd be good practice to use the same name/formatting as the error. What do you think of:

`test_workflow_dspace_client_raise_AuthenticationError`? 🤔

dsc/workflows/digitized_theses/workflow.py
```diff
@@ -0,0 +1,328 @@
...
+            # TODO: Custom error includes HTTPError message
+            raise exceptions.ItemMetadataNotFoundError from exception
+
+        # get nested `<marc:record>` element from SRU response
+        record_element = self._parse_record_from_sru_response(response.content)
+
+        # write an XML file to batch folder
+        s3_client.put_file(
+            bucket=self.s3_bucket,
+            key=f"{self.workflow_name}/{self.minted_batch_id}/{item_submission.item_identifier}.xml",
```
ghukill: Unsure how much the comments about about "minted" identifiers impacts this, but just noting more of these path constructions here.  To the degree these could lean on properties, I think the better.

dsc/workflows/digitized_theses/workflow.py
```diff
@@ -0,0 +1,328 @@
...
+            )
+
+        if len(dspace_objects) == 0:
+            return None
+
+        return dspace_objects[0]
+
+    def _move_batch_files_to_theses_subfolders(
+        self, item_submissions: list[ItemSubmission]
+    ) -> None:
```
ghukill, line 372: I think this could use a docstring as well.  While the workflow class level docstring may be the broad strokes, it feels like this method could be have detail about why certain items end up in certain folders.

dsc/workflows/digitized_theses/workflow.py
```diff
@@ -0,0 +1,328 @@
...
+                    f"Cannot move files associated with item_identifier={item_submission.item_identifier} "  # noqa: E501
+                    "because thesis type (new vs. replacement) is unknown"
+                )
+                continue
+
+            logger.debug(
+                "Moving files associated with "
+                f"item_identifier={item_submission.item_identifier} to '{prefix}/'"
+            )
+            for file in s3_client.files_iter(
```
ghukill: Noting that _again_ here, it gets confusing if we're copying files from the original S3 bucket where Jenn / digitized theses team uploaded items, or have they already been moved into the batch folder, but now we're moving them _within_ the batch folder?  

I think some of those early comments may help with this a bit, as perhaps we won't be juggling "minted_batch_id" anymore.

ghukill (commented):
I thought it better to submit my review now, with some requests for comments + docstrings, and some possible suggestions for the "minted" batch id language and logic. 

If it were possible to drop the word "mint" entirely, except maybe _once_ when some code "mints the batch id going forward for this batch", that feels like it could be a big win.  This might remove some confusion when scanning code if we're:

1. copying files from digi theses S3 workspace bucket to batch folder in DSC bucket 
2. moving files _within_ the batch folder

Which all brings me to a question.... is there a universe where this workflow should be preparing the batch locally -- e.g. in a python temp directory -- and then uploading the final, perfectly formed batch with it's new "minted" batch id to the DSC bucket?

If you went that path, if something went horribly wrong along the way, there isn't cruft in the DSC bucket.  Nothing got uploaded, because nothing was really ready enough.  And instead of moving files around in S3, which is kind of tricky, you're literally just moving around files in a python temp directory.  For this app, that'd be inside the ECS task Docker container, which is just fine.

Not a hard requirement / request to do this file downloading, renaming, and moving around locally _then_ uploading to DSC, but maybe worth a consideration whilst considering the comments about "minted" batch ids and all fairly repetitive S3 path constructions throughout.

All that said, this is some **complicated** stuff, really good foundation here conceptually I think.

dsc/workflows/digitized_theses/workflow.py
jonavellecuerdo: Updated to capitalize members of `StrEnum`!

dsc/workflows/digitized_theses/workflow.py
jonavellecuerdo: Updated docstring.

dsc/db/models.py
ehanson8: Correct, DynamoDB will write whatever but we can talk about whether consistency in useful or not in the ticket

tests/workflows/digitized_theses/test_workflow.py
ehanson8: Ah, I see, not a big deal either way, do what you feel!

dsc/workflows/digitized_theses/workflow.py
jonavellecuerdo: This has been updated.

jonavellecuerdo:
@ehanson8 @ghukill I just pushed up two commits--the former comprising of smaller, minor updates and the latter being more involved. In this description, I will be summarizing the changes in the second commit, which emerged from three key things/questions @ghukill pointed out:

1. The workflow module could use more docstrings.
   * Please view [the workflow module](https://github.com/MITLibraries/dspace-submission-composer/blob/IN-1626-digitized-theses-workflow-create/dsc/workflows/digitized_theses/workflow.py).
   * I plan to update the class docstring as I implement methods for each of the DSO steps (i.e., the next PR will include text for 'Queue a batch for ingest'
2. The workflow now [prepares the batch in a temporary directory on local disk ](https://github.com/MITLibraries/dspace-submission-composer/blob/IN-1626-digitized-theses-workflow-create/dsc/workflows/digitized_theses/workflow.py#L211-L221) and [once the batch is ready, it creates the batch folder in the DSC S3 bucket using `run_aws_cli_sync`
](https://github.com/MITLibraries/dspace-submission-composer/blob/IN-1626-digitized-theses-workflow-create/dsc/workflows/digitized_theses/workflow.py#L273-L283)
3. The workflow now moves item submission assets into [nested prefixes named with the item identifier **under** the `[replacement|new|skipped]-theses` prefixes](https://github.com/MITLibraries/dspace-submission-composer/blob/IN-1626-digitized-theses-workflow-create/dsc/workflows/digitized_theses/workflow.py#L373-L419). 
5. This workflow is the first time we want to use `synced=True` to avoid unnecessary API calls when creating a batch _in Prod_ via _syncing data from Stage_. 
   * The [`prepare_batch()` method now has branching logic for when `synced=True`](https://github.com/MITLibraries/dspace-submission-composer/blob/IN-1626-digitized-theses-workflow-create/dsc/workflows/digitized_theses/workflow.py#L119-L136).
   * When `synced=True`, this means:
      * A ready batched folder already exists (in Prod). 
   * All `prepare_batch()` needs to do is derive `ItemSubmission`'s given the contents of the ready batch folder to create/materialize the item submissions in DynamoDB.
   * ⚠️ : Requesting reviewers to wait on tests for [DigitizedTheses._get_item_submissions_from_synced_batch()](https://github.com/MITLibraries/dspace-submission-composer/blob/IN-1626-digitized-theses-workflow-create/dsc/workflows/digitized_theses/workflow.py#L138-L186). 
      * The way the unit tests mock DynamoDB calls is a bit messy and would like some time to ponder this, but wanting to prioritize implementation of DSO steps.

---
**Screenshots showing new structure of prepared batch folder**

image

image

image



Please let me know if you have any questions!
**Note:** I resolved Graham's comments in an attempt to summarize the key points that were shared in **this** comment.

dsc/workflows/digitized_theses/workflow.py
```diff
@@ -0,0 +1,468 @@
...
+    synced to a similarly named batch folder in the DSC S3 bucket and
+    ItemSubmission's are recorded in DynamoDB.
+
+    When this workflow creates a batch *with* syncing, this means an organized
+    batch folder already exists in the DSC S3 bucket in Prod, which is the
+    result of syncing data from Stage. This method will instead walk
+    the contents of the synced batch folder to create ItemSubmission's
+    that are recorded in DynamoDB.
+    ---
+    TBD
```
ghukill: This is a *great* docstring, thanks for adding.   Two 100% optional thoughts/comments:

1. Unsure if we need the "Create a batch" section per se?  I feel like you could just jump straight into that text.
2. What's the "TBD" at the end?  was this a placeholder for submit + finalize information when it's added?   See comment above.  I think ideally this docstring is kind of a single group of paragraphs, all one narrative, that wouldn't need that kind of sectioning.

dsc/workflows/digitized_theses/workflow.py
```diff
@@ -0,0 +1,468 @@
...
+            seen_item_identifiers.append(item_identifier)
+            item_submissions.append(
+                ItemSubmission(
+                    batch_id=self.batch_id,
+                    item_identifier=item_identifier,
+                    workflow_name=self.workflow_name,
+                    status=(
+                        ItemSubmissionStatus.CREATE_SUCCESS
+                        if theses_subfolder in ["replacement-theses", "new-theses"]
+                        else ItemSubmissionStatus.CREATE_SKIPPED
```
ghukill, line 178: It's extremely minor, not a requested change, but this _could_ be a standalone block before the `ItemSubmission` instantiation:

```python
status = (
    ItemSubmissionStatus.CREATE_SUCCESS
    if theses_subfolder in ["replacement-theses", "new-theses"]
    else ItemSubmissionStatus.CREATE_SKIPPED
)
item_submissions.append(
    ItemSubmission(
        batch_id=self.batch_id,
        item_identifier=item_identifier,
        workflow_name=self.workflow_name,
        status=status,
    )
)
```

dsc/workflows/digitized_theses/workflow.py
```diff
@@ -0,0 +1,468 @@
...
+
+        1. Download the MARC XML record from Alma and save to an XML file
+        2. Check if an item in DSpace already exists:
+            - If it exists and it is a student-submitted electronic thesis, skip
+            - If it exists and is not, mark as 'Replacement thesis'
+            - If it does not exist, mark as 'New thesis'
+        3. Organize the contents of the dated batch folder by thesis type
+        4. Sync the dated batch folder to the workflow folder in the DSC S3 bucket
+        5. Clean up the temporary directory
+        """
```
ghukill, line 201: Chef's kiss!  Another great docstring, thanks for adding.

Also, whilst here, thanks for entertaining doing this batch creation locally, "on disk".  I really think it makes sense doing this kind of complex file checking and arrangement locally, then sending to S3.

ghukill: On a re-read, I think this docstring could be a good place to explain that _only_ here do you modify the batch id.  It may not be necessary to really explain why -- becuase multiple DSC batch attempts may originate from a single folder in the workspace? -- but that once modified and set here, it's effectively set in Dynamo and future submit and finalize commands will get their batch ids from that.

Please feel free not to add if not helpful.

**Double update**: I clicked into the `_update_batch_id` method and see documentation there.    Works for me!

dsc/workflows/digitized_theses/workflow.py
```diff
@@ -0,0 +1,468 @@
...
+    def _move_batch_files_to_theses_subfolders(
+        self, item_submissions: list[ItemSubmission], batch_location: str
+    ) -> None:
+        """Move item submission assets in batch folder into theses subfolders.
+
+        This method will inspect the `status_details` for each item submission to
+        determine which theses subfolder (prefix) its content should be moved to.
+        At this point, `status_details` will contain "Replacement thesis",
+        "New thesis", or an exception message. If `status_details` contains an
+        exception message, contents are moved to a skipped-theses/ subfolder.
```
ghukill, line 379: I'm not crazy about this kind of fuzzy string matching/checking for classifying the item as new vs. replacement vs. skip/error, mostly for the possibility than an exception message from this statement which shows up a couple of times,

```python
item_submission.status_details = str(exception)
```

may somehow include words like "new" or "replacment" and get incorectly categorized.

Would it work to look for more exact phrases like `"Replacement thesis"` or `"New thesis"` in the status details?  maybe removing the `.lower()` as well?  Maybe that could add some additional safety without really changing the approach?  

Another thought, note that `ItemSubmission` class has these already on it:

```python
"""
...
2. Processing attributes: Temporary data used during submission processing
   but not persisted
...
"""

...
# processing attributes
source_metadata: dict[str, Any] | None = None
dspace_metadata: dict[str, Any] | None = None
bitstream_s3_uris: list[str] | None = None
metadata_s3_uri: str = ""
```

You _could_ update that class to have something like `.processing_meta` which is a freeform dictionary to note things on the `ItemSubmission` instance during processing.  Then, use that during sorting and categorizing.

Another option...

Instead of appending `ItemSubmission` instances into `item_submissions`, you could store them in a dictionary like:

```python
item_submissions = {
  "new":[...],
  "replacement":[...],
  "skip_and_error":[...],
}
```

If you pass that dictionary to `self._move_batch_files_to_theses_subfolders()` it of course could quite easily just slot them into the right folders.

Then, when ready to return the list you could do

```python
return (
  item_submissions["new"] +
  item_submissions["replacement"] +
  item_submissions["skip_and_error"]
)
```

Or this, which AI suggested and I wouldn't have thought of!

```python
from itertools import chain

return list(chain(
    item_submissions["new"],
    item_submissions["replacement"],
    item_submissions["skip_and_error"],
))
```

**TL/DR:** It really might be worth avoiding the fuzzy matching of "new" and "replacement" in the `ItemSubmission.status_details` as a sorting mechanism, given that we're storing raw exceptions; they could theoretically contain those words.  A bit of effort to avoid now, might avoid a _really_ sneaky bug in production later.

ghukill (changes requested):
Looking really good, nice work wrangling some complex mechanics.

I've opted for a "Request changes" 99% for the categorization of items based on their `status_details`.  For reasons noted in the comment, I'd lobby pretty strong to avoid the "new" or "replacment" fuzzy string matching.  Open to pushback and more discussion, but this feels like something to really kind of iron out.

Otherwise, basically an approval else-wise!

dsc/workflows/digitized_theses/workflow.py
jonavellecuerdo: 1. Will remove section headers!
2. It was intended to be a placeholder. 😅 Will remove and replace with a `TODO` instead.

dsc/workflows/digitized_theses/workflow.py
jonavellecuerdo: I will update the conditions to check for exact matches of "New thesis" and "Replacement thesis". I will hold off on additional changes to the `ItemSubmission` class but will continue to keep them in mind.

dsc/workflows/digitized_theses/workflow.py
jonavellecuerdo: Hmm just sharing that I've gone back-and-forth over creating single-use variables vs. direct assignment 😅 , and I haven't yet formed a "rule" to consistently follow. Curious about yours and @ehanson8 considerations when deciding what approach to take?

ghukill: I believe that it's highly variable (pun intended) and not particularly rule based.  

In this scenario, I think this logic,

```python
ItemSubmissionStatus.CREATE_SUCCESS
if theses_subfolder in ["replacement-theses", "new-theses"]
else ItemSubmissionStatus.CREATE_SKIPPED
```

is particularly important to this workflow, and it feels like it gets lost in the `ItemSubmission` init.  To me, it's as simple as that. 

But as originally noted, 100% optional!  If you like what you've got currently, by all means.  Pros and cons on either side.

ghukill (approved):
Good stuff, solid movement forward.

ehanson8 (approved):

PR #241, opened 2026-05-27
Implement submit methods for digitized theses workflow

jonavellecuerdo:
The digitized theses workflow needs to send submission messages to DSS based on the thesis type ('New thesis' or 'Replacement thesis'). If a thesis is a replacement, DSS should execute an 'update' operation to replace the bitstream in the corresponding DSpace item.

Note(s):
* Taking on the advice of reviewers in the previous PR, I made sure to add detailed docstrings for new methods! I recommend reviewers read these carefully to understand the general flow of the `submit` process for this workflow.
* Given that `DigitizedTheses` required its own implementation of `submit_items()`, I figured this was a good opportunity to explore the creation of DSpace metadata that wasn't reliant on a metadata mapping file (see [`ItemSubmission.create_dspace_metadata_without_mapping`](https://github.com/MITLibraries/dspace-submission-composer/blob/IN-1626-digitized-theses-workflow-submit/dsc/item_submission.py#L325-L364)).

For now, a review of passing unit tests ([workflow-specific unit tests related to submit](https://github.com/MITLibraries/dspace-submission-composer/pull/241/changes#diff-2b34f3f7b9c6e04a94bf0855eb0a73c239e54098f467b2d3fb1f9c5a243a87a7R256-R412) + [updated unit tests for `SQSClient`](https://github.com/MITLibraries/dspace-submission-composer/pull/241/changes#diff-b4923bdc63319fa5b2ecf5c183abe8da5c590ee31399b4c51884b44b66c898cdR66-R89)) is sufficient.

NO

NO

### Code review


dsc/utils/aws/sqs.py
```diff
@@ -78,9 +80,12 @@ def create_dss_message_body(
         Args:
             submission_system: The system where the submission is uploaded
             (e.g. DSpace@MIT).
-            collection_handle: The handle of collection where the submission is uploaded.
             metadata_s3_uri: The S3 URI for the metadata JSON file.
             bitstream_s3_uris: The S3 URIs for the submission's bitstreams.
+            operation: The operation to perform for an item, 'create' or 'update'.
+            Defaults to 'create'.
+            collection_handle: The handle for the community in which an item is created.
```
ehanson8: `community` > `collection`

dsc/utils/aws/sqs.py
```diff
@@ -78,9 +80,12 @@ def create_dss_message_body(
...
+            item_handle: The handle for an item to update.
```
ehanson8: `The handle of an item to be updated`?

dsc/workflows/base/workflow.py
```diff
@@ -340,7 +341,7 @@ def submit_items(self, collection_handle: str | None = None) -> list:
         )
         return items
 
-    def _get_item_collection_handle(self) -> str:
+    def _get_item_collection_handle(self, item_metadata: dict) -> str:
```
ehanson8, line 344: It would be more verbose but you consider might considered `_get_collection_handle_from_metadata` to more accurately describe this method

dsc/workflows/digitized_theses/transformer.py
```diff
@@ -140,13 +140,9 @@ def transform(cls, source_metadata: str | bytes) -> dict:
             source_metadata: A complete MARC21/XML document of metadata from
             Alma SRU as a byte string.
         """
-        root = etree.fromstring(source_metadata)
-        record = cls._find_record(root)
-        if record is None:
-            raise ValueError("No <marc:record> element found in input.")
```
ehanson8, line 146: Is this scenario not expected anymore or did the check shift somewhere else?

dsc/workflows/digitized_theses/workflow.py
```diff
@@ -26,6 +29,14 @@
 CONFIG = Config()
 logger = logging.getLogger(__name__)
 
+# TODO: Update collection handles for test instance once ready
+MIT_THESES_COLLECTION_HANDLES = {
+    "Bachelor": {"dev": "1721.1/test", "prod": "1721.1/131024"},
+    "Engineer": {"dev": "1721.1/test", "prod": "1721.1/131023"},
+    "Master": {"dev": "1721.1/test", "prod": "1721.1/131023"},
+    "Doctoral": {"dev": "1721.1/test", "prod": "1721.1/131022"},
```
ehanson8, line 37: Hardcoding like this makes me a little twitchy even if they're handles that shouldn't change, could Parameter Store be a good place for these values?

dsc/workflows/digitized_theses/workflow.py
```diff
@@ -435,6 +454,226 @@ def _is_replacement_thesis(item: DSpaceItem) -> bool:
 
         return True
 
+    def submit_items(self, collection_handle: str | None = None) -> list:
```
ehanson8: Should it be `_collection_handle` since the method doesn't actually use it?

dsc/workflows/digitized_theses/workflow.py
```diff
@@ -435,6 +454,226 @@ def _is_replacement_thesis(item: DSpaceItem) -> bool:
...
+                    batch_path=self.batch_path,
+                )
+                item_submission.bitstream_s3_uris = manifest[
+                    item_submission.item_identifier
+                ]["bitstream_files"]
+
+                # get collection handle based on mit.theses.degree
+                item_submission.collection_handle = (
+                    collection_handle or self._get_item_collection_handle(item_metadata)
+                )
```
ehanson8: Shouldn't this be in the `else` block for new theses? It doesn't look like it's used for replacements

dsc/item_submission.py
```diff
@@ -352,7 +408,10 @@ def send_submission_message(
             output_queue: The SQS output queue used for retrieving result messages.
             submission_system: The system where the submission is uploaded
             (e.g. DSpace@MIT).
-            collection_handle: The handle of collection where the submission is uploaded.
+            operation: The operation to perform for an item, 'create' or 'update'.
+            Defaults to 'create'.
+            collection_handle: The handle for the community in which an item is created.
+            item_handle: The handle for an item to update.
```
ehanson8: Same suggestions as earlier comment

ehanson8 (commented):
Looking good, a few questions!

dsc/utils/aws/sqs.py
```diff
@@ -65,9 +65,11 @@ def create_dss_message_attributes(
     @staticmethod
     def create_dss_message_body(
         submission_system: str,
-        collection_handle: str,
         metadata_s3_uri: str,
         bitstream_s3_uris: list[str],
+        operation: Literal["create", "update"] | None = "create",
+        collection_handle: str | None = None,
+        item_handle: str | None = None,
```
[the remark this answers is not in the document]
jonavellecuerdo: All invocations of `create_dss_message_body` use keyword args.

dsc/workflows/base/workflow.py
jonavellecuerdo: If it's okay, I'd like to keep the name as-is. My thinking is the argument already being `item_metadata` communicates where this info is derived from. 🤔

dsc/workflows/digitized_theses/transformer.py
jonavellecuerdo: It's not expected anymore. 🤔 When DSC writes Alma MARC XML record to a file, it retrieves the nested `<marc:record>` element from the SRU response.

dsc/workflows/digitized_theses/workflow.py
jonavellecuerdo: Adding this to a list of items to discuss with @cabutlermit on how best to store such configs! Are you thinking we'd use the SSM client to access the value of these parameters? 

For the purpose of this PR, I am hoping we can proceed with this, knowing that it will be discussed as part of: https://github.com/MITLibraries/dspace-submission-composer/pull/241#discussion_r3313619518

dsc/workflows/digitized_theses/workflow.py
jonavellecuerdo: Great catch! I moved it to the `else` block.

dsc/workflows/digitized_theses/workflow.py
ehanson8: Fine by me!

ehanson8 (approved):
Changes look great, approved!

dsc/workflows/digitized_theses/workflow.py
cabutlermit: These really shouldn't be hardcoded. But, fine to do on this merge as long as there is a fix coming in the near future to move these out of the code and into a parameter. We can discuss exactly how to format this parameter in a separate conversation. Also: where is the "stage" handle?

jonavellecuerdo: @cabutlermit Thanks for providing input! The app contains logic so that when `CONFIG.workspace != "prod"` will default to the `dev` handle (i.e., the handle in the **test instance** of MIT Open Scholarship / DSpace@MIT).

dsc/item_submission.py
```diff
@@ -218,6 +218,12 @@ def ready_to_submit(self) -> bool:
                                       item_identifier=self.item_identifier)
                     } " " already submitted, skipping submission"
                 )
+            case ItemSubmissionStatus.CREATE_FAILED | ItemSubmissionStatus.CREATE_SKIPPED:
+                logger.info(
+                    f"Record " f"{ITEM_SUBMISSION_LOG_STR.format(batch_id=self.batch_id,
+                                      item_identifier=self.item_identifier)
+                    } " " submission assets invalid, skipping submission"
```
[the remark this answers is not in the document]
jonavellecuerdo: This is a known error and will be updated in a follow-up PR.

dsc/workflows/digitized_theses/workflow.py
cabutlermit: got it. in the future, when we move this to a parameter, we should be consistent with naming between key name in the parameter and the in-practice name for the "test" instance of the server (do we call that server "test" or "stage" or "dev" or "bob's your uncle" 😆 ).

dsc/workflows/digitized_theses/workflow.py
```diff
@@ -435,6 +454,226 @@ def _is_replacement_thesis(item: DSpaceItem) -> bool:
...
+                if (
+                    manifest[item_submission.item_identifier]["thesis_type"]
+                    == "Replacement thesis"
+                ):
+                    response = item_submission.send_submission_message(
+                        submission_source=self.workflow_name,
+                        output_queue=self.output_queue,
+                        submission_system=self.submission_system,
+                        operation="update",
+                        item_handle=item_submission.dspace_handle,
```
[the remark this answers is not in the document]
jonavellecuerdo: @ehanson8 Just flagging this great catch by Copilot! I think the required changes will require some code reorganization around DSpace client calls. I created a separate ticket: https://mitlibraries.atlassian.net/browse/IN-1756.

**Note:** There was some discussion with Sadie and Jenn re: skipping "test runs" in Stage and doing all the work in Prod so the 'sync batches from Stage' may not be used at all. Nonetheless, I created the ticket to address the issue.

PR #245, opened 2026-06-03
In 1753 align create statuses across workflows

jonavellecuerdo:
Statuses for item submission post batch creation now comprise of "create_success|failed|skipped", not just "batch_created". This change is to support workflows that require writing item submissions to the DynamoDB table even if an error occurs during batch creation. As a reminder, the `opencourseware` and `sccs` workflows will only write item submissions to DynamoDB when no errors occur during the batch creation step, whereas the `digitized-theses` workflow allows item submissions with status `CREATE_FAILED|CREATE_SKIPPED`. 

The following updates were applied:
* [Remove default `status` assignment of `CREATE_SUCCESS` from base `Workflow.create_batch()` method
](https://github.com/MITLibraries/dspace-submission-composer/pull/245/changes/4150147cf83251548f170ab468cba57e900e6dac#diff-7f2dbbca549d2a37648633455a3a61bdeeedb3aa760ce59172269f2122341b44L257-L258)
* Update [`OpenCourseWare`](https://github.com/MITLibraries/dspace-submission-composer/pull/245/changes/4150147cf83251548f170ab468cba57e900e6dac#diff-7c98a93b34b4e4af217311e6e7c97eb79a491377671e64200bfbbfe9d806214bR129) and [`SCCS`](https://github.com/MITLibraries/dspace-submission-composer/pull/245/changes/4150147cf83251548f170ab468cba57e900e6dac#diff-751224c9144d2dc37bf5cb4afb7d0a632d02461c410cea27af9d14efa887487fR123) workflow to set statuses during `prepare_batch()` method

NOTE: The second commit in this PR resolves an issue introduced by the reporting module updates, which unintentionally removed the creation of an errors CSV file, which is still required for the `opencourseware` and `sccs` workflows, which do not write records to DynamoDB in the event of a batch creation error. The proposed solution in the second commit is intended to be minimal, and it's worth continuing to reflect on this difference when considering future enhancements to reporting.

For now, a review of code changes is sufficient.

NO

NO

- https://mitlibraries.atlassian.net/browse/IN-1753

### Code review

### DIFF
--- a/dsc/cli.py
+++ b/dsc/cli.py
@@ -133,13 +133,20 @@ def create(
             )
             ctx.exit(exception.exit_code)
 
+    # init list of batch creation errors
+    batch_creation_errors: list = []
     try:
         workflow.create_batch(synced=sync_data)
-    except BatchCreationFailedError:
+    except BatchCreationFailedError as exception:
         logger.error("Failed to create batch")  # noqa: TRY400
+        batch_creation_errors.extend(exception.errors)
 
     if email_recipients:
-        workflow.send_report(step="create", email_recipients=email_recipients.split(","))
+        workflow.send_report(
+            step="create",
+            email_recipients=email_recipients.split(","),
+            errors=batch_creation_errors,
+        )
 
 
 # data sync command
--- a/dsc/reports/base.py
+++ b/dsc/reports/base.py
@@ -40,9 +40,10 @@ class Report(ABC):
         ),
     )
 
-    def __init__(self, workflow_name: str, batch_id: str):
+    def __init__(self, workflow_name: str, batch_id: str, errors: list | None = None):
         self.workflow_name = workflow_name
         self.batch_id = batch_id
+        self.errors = errors
         self.report_date = datetime.now(tz=UTC).strftime("%Y%m%dT%H%M%SZ")
 
         # configure environment for loading jinja templates
@@ -55,8 +56,10 @@ class Report(ABC):
         self._item_submissions: list[ItemSubmission] | None = None
 
     @classmethod
-    def load(cls, workflow_name: str, batch_id: str) -> "Report":
-        return cls(workflow_name=workflow_name, batch_id=batch_id)
+    def load(
+        cls, workflow_name: str, batch_id: str, errors: list | None = None
+    ) -> "Report":
+        return cls(workflow_name=workflow_name, batch_id=batch_id, errors=errors)
 
     @property
     @abstractmethod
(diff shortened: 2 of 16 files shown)


tests/conftest.py
```diff
@@ -211,7 +213,7 @@ def metadata_mapping():
 
 
 @pytest.fixture
-def mocked_item_submission_db(config_instance):
+def mock_item_submission_db(config_instance):
```
ehanson8, line 216: We should probably be consistent about `mock_` vs. `mocked_` in this repo (and in general) but that's likely outside the scope of this PR

dsc/reports/base.py
```diff
@@ -40,7 +40,7 @@ class Report(ABC):
         ),
     )
 
-    def __init__(self, workflow_name: str, batch_id: str):
+    def __init__(self, workflow_name: str, batch_id: str, **kwargs: list):
```
ehanson8: Could this be an optional `errors` param? `kwargs` makes me twitchy since it isn't clear what's expected. You are only talking about `errors` in this implementation so `kwargs` seems unnecessarily broad

ehanson8 (commented):
Question about `kwargs`

tests/conftest.py
jonavellecuerdo: My goal is to change all `mocked_` to `mock_` when revising unit tests to align with examples I usually see when reading up on mocking, but yes, all will be updated eventually!

dsc/reports/base.py
jonavellecuerdo: Thanks for the conversation around the use of `errors`.

dsc/cli.py
```diff
@@ -135,11 +135,14 @@ def create(
...
-    except BatchCreationFailedError:
+    except BatchCreationFailedError as exception:
         logger.error("Failed to create batch")  # noqa: TRY400
+        errors = exception.errors
 
     if email_recipients:
-        workflow.send_report(step="create", email_recipients=email_recipients.split(","))
+        workflow.send_report(
+            step="create", email_recipients=email_recipients.split(","), errors=errors
+        )
```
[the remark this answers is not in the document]
jonavellecuerdo: This was a good catch. It does result in the following code: 

```python
 # init list of batch creation errors
    batch_creation_errors: list = []
    try:
        workflow.create_batch(synced=sync_data)
    except BatchCreationFailedError as exception:
        logger.error("Failed to create batch")  # noqa: TRY400
        batch_creation_errors.extend(exception.errors)

    if email_recipients:
        workflow.send_report(
            step="create",
            email_recipients=email_recipients.split(","),
            errors=batch_creation_errors,
        )
```

Not liking having to init a variable with "errors" as one of the first lines of code to execute, but moving forward with this for now!

Tagging @ehanson8 for awareness!

ehanson8 (approved):
Great work!

PR #257, opened 2026-08-04
Remap department values to dc.relation.orgunit

jonavellecuerdo:
All collections will be using the dc.relation.orgunit metadata field instead of dc.contributor.department moving forward. All workflows that currently derive dc.contributor.department need to be updated.

This is a simple change for which passing unit tests are sufficient! In addition, the recurring `403 Forbidden Errors` with the test instance of MIT Open Scholarship still need to be addressed before we can run any ingests to it. 

YES - Ran `make update`; required addressing some new Ruff rules: [CPY001](https://docs.astral.sh/ruff/rules/missing-copyright-notice/), [PLR0917](https://docs.astral.sh/ruff/rules/too-many-positional-arguments/), [ISC004](https://docs.astral.sh/ruff/rules/implicit-string-concatenation-in-collection-literal/)

NO

- https://mitlibraries.atlassian.net/browse/IN-1801

### Code review

### DIFF
--- a/dsc/utils/aws/sqs.py
+++ b/dsc/utils/aws/sqs.py
@@ -63,7 +63,7 @@ class SQSClient:
         }
 
     @staticmethod
-    def create_dss_message_body(
+    def create_dss_message_body(  # noqa: PLR0917
         submission_system: str,
         metadata_s3_uri: str,
         bitstream_s3_uris: list[str],
--- a/dsc/item_submission.py
+++ b/dsc/item_submission.py
@@ -437,7 +437,7 @@ class ItemSubmission:
         logger.info(f"Metadata uploaded to S3: {metadata_s3_uri}")
         self.metadata_s3_uri = metadata_s3_uri
 
-    def send_submission_message(
+    def send_submission_message(  # noqa: PLR0917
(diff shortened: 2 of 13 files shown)


dsc/utils/aws/sqs.py
```diff
@@ -63,7 +63,7 @@ def create_dss_message_attributes(
         }
 
     @staticmethod
-    def create_dss_message_body(
+    def create_dss_message_body(  # noqa: PLR0917
```
ehanson8, line 66: Optional: I added this to `pyproject.toml` as well since we have ignored similar `PLR` rules

ehanson8 (approved):
Looks great! One optional suggestion

dsc/utils/aws/sqs.py
jonavellecuerdo: Good to know! I'll leave as-is for now. It did make me wonder if we should be more explicit about using kwargs for methods like these. 🤔 There have been a few instances where I reordered the arguments but unit tests fail because I got the wrong position of args (i.e., would've been easier to catch required changes with kwargs).

